PR #5, opened 2016-10-30
new PCS format

Krxsy:
Added pcs_new.py as new pcs format for the latest SMAC release (thus OrdinalHyperparameter, two conditions (less/greater than) and some relevant unit tests).

### DIFF
--- a/ConfigSpace/conditions.py
+++ b/ConfigSpace/conditions.py
@@ -181,11 +181,13 @@ class AbstractConjunction(ConditionComponent):
 class EqualsCondition(AbstractCondition):
     def __init__(self, child, parent, value):
         super(EqualsCondition, self).__init__(child, parent)
+        '''
         if not parent.is_legal(value):
             raise ValueError("Hyperparameter '%s' is "
                              "conditional on the illegal value '%s' of "
                              "its parent hyperparameter '%s'" %
                              (child.name, value, parent.name))
+        '''
         self.value = value
 
     def __repr__(self):
@@ -199,11 +201,14 @@ class EqualsCondition(AbstractCondition):
 class NotEqualsCondition(AbstractCondition):
     def __init__(self, child, parent, value):
         super(NotEqualsCondition, self).__init__(child, parent)
+        '''
         if not parent.is_legal(value):
             raise ValueError("Hyperparameter '%s' is "
                              "conditional on the illegal value '%s' of "
                              "its parent hyperparameter '%s'" %
                              (child.name, value, parent.name))
+                             
+        '''
         self.value = value
 
     def __repr__(self):
@@ -212,7 +217,42 @@ class NotEqualsCondition(AbstractCondition):
 
     def _evaluate(self, value):
         return value != self.value
-
+        
+class LessThanCondition(AbstractCondition):
+    def __init__(self, child, parent, value):
+        super(LessThanCondition, self).__init__(child, parent)
+        '''
+        if not parent.is_legal(value):
+            raise ValueError("Hyperparameter '%s' is "
+                             "conditional on the illegal value '%s' of "
+                             "its parent hyperparameter '%s'" %
+                             (child.name, value, parent.name))
+        '''
+        self.value = value
+        
+    def __repr__(self):
+        return "%s | %s < %s" % (self.child.name, self.parent.name,
+                                  repr(self.value))
+    def _evaluate(self, value):
+        return value < self.value
+    
+class GreaterThanCondition(AbstractCondition):
+    def __init__(self, child, parent, value):
+        super(GreaterThanCondition, self).__init__(child, parent)
+        '''
+        if not parent.is_legal(value):
+            raise ValueError("Hyperparameter '%s' is "
+                             "conditional on the illegal value '%s' of "
+                             "its parent hyperparameter '%s'" %
+                             (child.name, value, parent.name))
+        '''
+        self.value = value
+        
+    def __repr__(self):
+        return "%s | %s > %s" % (self.child.name, self.parent.name,
+                                  repr(self.value))
+    def _evaluate(self, value):
+        return value > self.value
 
 class InCondition(AbstractCondition):
     def __init__(self, child, parent, values):
--- a/ConfigSpace/conditions.py
+++ b/ConfigSpace/conditions.py
@@ -181,13 +181,11 @@ class AbstractConjunction(ConditionComponent):
 class EqualsCondition(AbstractCondition):
     def __init__(self, child, parent, value):
         super(EqualsCondition, self).__init__(child, parent)
-        '''
         if not parent.is_legal(value):
             raise ValueError("Hyperparameter '%s' is "
                              "conditional on the illegal value '%s' of "
                              "its parent hyperparameter '%s'" %
                              (child.name, value, parent.name))
-        '''
         self.value = value
 
     def __repr__(self):
@@ -201,14 +199,11 @@ class EqualsCondition(AbstractCondition):
 class NotEqualsCondition(AbstractCondition):
     def __init__(self, child, parent, value):
         super(NotEqualsCondition, self).__init__(child, parent)
-        '''
         if not parent.is_legal(value):
             raise ValueError("Hyperparameter '%s' is "
                              "conditional on the illegal value '%s' of "
                              "its parent hyperparameter '%s'" %
                              (child.name, value, parent.name))
-                             
-        '''
         self.value = value
 
     def __repr__(self):
@@ -221,13 +216,11 @@ class NotEqualsCondition(AbstractCondition):
 class LessThanCondition(AbstractCondition):
     def __init__(self, child, parent, value):
         super(LessThanCondition, self).__init__(child, parent)
-        '''
         if not parent.is_legal(value):
             raise ValueError("Hyperparameter '%s' is "
                              "conditional on the illegal value '%s' of "
                              "its parent hyperparameter '%s'" %
                              (child.name, value, parent.name))
-        '''
         self.value = value
         
     def __repr__(self):
@@ -239,13 +232,11 @@ class LessThanCondition(AbstractCondition):
 class GreaterThanCondition(AbstractCondition):
     def __init__(self, child, parent, value):
         super(GreaterThanCondition, self).__init__(child, parent)
-        '''
         if not parent.is_legal(value):
             raise ValueError("Hyperparameter '%s' is "
                              "conditional on the illegal value '%s' of "
                              "its parent hyperparameter '%s'" %
                              (child.name, value, parent.name))
-        '''
         self.value = value
         
     def __repr__(self):
@@ -285,10 +276,12 @@ class AndConjunction(AbstractConjunction):
 
     def __repr__(self):
         retval = six.StringIO()
+        retval.write("(")
         for idx, component in enumerate(self.components):
             retval.write(str(component))
             if idx < len(self.components) - 1:
                 retval.write(" && ")
+        retval.write(")")
         return retval.getvalue()
 
     def _evaluate(self, evaluations):
@@ -304,10 +297,12 @@ class OrConjunction(AbstractConjunction):
 
     def __repr__(self):
         retval = six.StringIO()
+        retval.write("(")
         for idx, component in enumerate(self.components):
             retval.write(str(component))
             if idx < len(self.components) - 1:
                 retval.write(" || ")
+        retval.write(")")
         return retval.getvalue()
 
     def _evaluate(self, evaluations):
--- a/ConfigSpace/conditions.py
+++ b/ConfigSpace/conditions.py
@@ -285,12 +285,10 @@ class AndConjunction(AbstractConjunction):
 
     def __repr__(self):
         retval = six.StringIO()
-        retval.write("(")
         for idx, component in enumerate(self.components):
             retval.write(str(component))
             if idx < len(self.components) - 1:
                 retval.write(" && ")
-        retval.write(")")
         return retval.getvalue()
 
     def _evaluate(self, evaluations):
@@ -306,12 +304,10 @@ class OrConjunction(AbstractConjunction):
 
     def __repr__(self):
         retval = six.StringIO()
-        retval.write("(")
         for idx, component in enumerate(self.components):
             retval.write(str(component))
             if idx < len(self.components) - 1:
                 retval.write(" || ")
-        retval.write(")")
         return retval.getvalue()
 
     def _evaluate(self, evaluations):
--- a/ConfigSpace/hyperparameters.py
+++ b/ConfigSpace/hyperparameters.py
@@ -29,6 +29,8 @@
 from abc import ABCMeta, abstractmethod
 import warnings
 
+from collections import OrderedDict
+
 import numpy as np
 import six
 
@@ -729,6 +731,12 @@ class OrdinalHyperparameter(Hyperparameter):
         self.sequence= sequence
         self._num_elements = len(sequence)
         self.default = self.check_default(default)
+        
+        self.value_dict = OrderedDict()
+        counter = 1
+        for element in self.sequence:
+            self.value_dict[element] = counter
+            counter += 1
 
     def __repr__(self):
         repr_str = six.StringIO()
@@ -756,6 +764,22 @@ class OrdinalHyperparameter(Hyperparameter):
             return default
         else:
             raise ValueError("Illegal default value %s" % str(default))
+            
+    def get_seq_order(self):
+        return np.arange(1,self._num_elements+1)
+        
+    def get_order(self, value):
+        return self.value_dict[value]
+        
+    def get_value(self, idx):
+        return list(self.value_dict.keys())[list(self.value_dict.values()).index(idx)]
+            
+    def check_order(self,value):
+        idx = self.get_order(value)
+        if idx < self.sequence[idx+1]:
+            return True
+        else:
+            return False
 
     def _sample(self, rs, size=None):
         return rs.randint(0, self._num_elements, size=size)
--- a/ConfigSpace/hyperparameters.py
+++ b/ConfigSpace/hyperparameters.py
@@ -722,4 +722,96 @@ class CategoricalHyperparameter(Hyperparameter):
                     neighbors.append(candidate)
 
         return neighbors
+        
+class OrdinalHyperparameter(Hyperparameter):
+    def __init__(self, name, sequence, default=None):
+        super(OrdinalHyperparameter, self).__init__(name)
+        self.sequence= sequence
+        self._num_elements = len(sequence)
+        self.default = self.check_default(default)
+
+    def __repr__(self):
+        repr_str = six.StringIO()
+        repr_str.write("%s, Type: Ordinal, Sequence: {" % (self.name))
+        for idx, seq in enumerate(self.sequence):
+            repr_str.write(str(seq))
+            if idx < len(self.sequence) - 1:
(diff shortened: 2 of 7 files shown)


ConfigSpace/conditions.py
```diff
@@ -181,11 +181,13 @@ def _evaluate(self, evaluations):
 class EqualsCondition(AbstractCondition):
     def __init__(self, child, parent, value):
         super(EqualsCondition, self).__init__(child, parent)
+        '''
```
mfeurer: Why is this check removed?

ConfigSpace/conditions.py
```diff
@@ -199,11 +201,14 @@ def _evaluate(self, value):
 class NotEqualsCondition(AbstractCondition):
     def __init__(self, child, parent, value):
         super(NotEqualsCondition, self).__init__(child, parent)
+        '''
```
mfeurer: Why is this check removed?

ConfigSpace/conditions.py
```diff
@@ -212,7 +217,42 @@ def __repr__(self):
 
     def _evaluate(self, value):
         return value != self.value
-
+        
+class LessThanCondition(AbstractCondition):
+    def __init__(self, child, parent, value):
+        super(LessThanCondition, self).__init__(child, parent)
+        '''
```
mfeurer: Why is this check deactivated?

ConfigSpace/conditions.py
```diff
@@ -212,7 +217,42 @@ def __repr__(self):
...
+        return "%s | %s < %s" % (self.child.name, self.parent.name,
+                                  repr(self.value))
+    def _evaluate(self, value):
+        return value < self.value
+    
+class GreaterThanCondition(AbstractCondition):
+    def __init__(self, child, parent, value):
+        super(GreaterThanCondition, self).__init__(child, parent)
+        '''
+        if not parent.is_legal(value):
```
mfeurer: Why is this check deactivated?

ConfigSpace/conditions.py
```diff
@@ -245,12 +285,10 @@ def __init__(self, *args):
 
     def __repr__(self):
         retval = six.StringIO()
```
mfeurer: Why is this bracket removed? This is only used for cosmetic reasons here

ConfigSpace/conditions.py
```diff
@@ -266,12 +304,10 @@ def __init__(self, *args):
...
-        retval.write("(")
```
mfeurer: Why is this bracket removed? This is only used for cosmetic reasons here

ConfigSpace/hyperparameters.py
```diff
@@ -722,4 +724,167 @@ def get_neighbors(self, value, rs, number=np.inf, transform=False):
...
+            self.value_dict[element] = counter
+            counter += 1
+
+    def __repr__(self):        
+        """
+        writes out the parameter definition
+        """
+        repr_str = six.StringIO()
+        repr_str.write("%s, Type: Ordinal, Sequence: {" % (self.name))
+        for idx, seq in enumerate(self.sequence):
```
mfeurer: How about `', '.join([self.sequence.keys()])`?

ConfigSpace/hyperparameters.py
```diff
@@ -722,4 +724,167 @@ def get_neighbors(self, value, rs, number=np.inf, transform=False):
...
+        checks if there are neighbors or we're only dealing with an 
+        one-element sequence
+        """
+        return len(self.sequence) > 1
+
+    def get_num_neighbors(self):
+        """
+        returns the number of existing neighbors in the sequence
+        """
+        return len(self.sequence) - 1
```
mfeurer: I think for an ordinal hyperparameter, there are only two neighbors at most, namely the smaller and the larger one.

ConfigSpace/hyperparameters.py
```diff
@@ -722,4 +724,167 @@ def get_neighbors(self, value, rs, number=np.inf, transform=False):
...
+        """
+        returns the neighbors of a given value
+        """
+        neighbors = []
+        if number < len(self.sequence):
+            while len(neighbors) < number:
+                rejected = True
+                index = int(value)
+                while rejected:
+                    neighbor_idx = rs.randint(0, self._num_elements)
```
mfeurer: This needs to be adjusted to account for the lower number of neighbors of an ordinal hyperparameter.

ConfigSpace/io/pcs_new.py
```diff
@@ -0,0 +1,475 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+__authors__ = ["Katharina Eggensperger", "Matthias Feurer"]
```
mfeurer: You should add yourself as the author here.

ConfigSpace/io/pcs_new.py
```diff
@@ -0,0 +1,475 @@
...
+    pyparsing.Optional(pyparsing.OneOrMore("," + pp_param_name + "=" + pp_numberorname)) + "}"
+
+
+def build_categorical(param):
+    cat_template = "%s categorical {%s} [%s]"
+    return cat_template % (param.name,
+                           ", ".join([str(value) for value in param.choices]),
+                           str(param.default))
+
+def build_constant(param):
```
mfeurer: I don't think there is a definition for a constant hyperparameter in the pcs definition, you need to encode it as a categorical.

ConfigSpace/io/pcs_new.py
```diff
@@ -0,0 +1,475 @@
...
+    for clause in forbidden:
+        param_list = pp_forbidden_clause.parseString(clause)
+        tmp_list = []
+        clause_list = []
+        for value in param_list[1:]:
+            if len(tmp_list) < 3:
+                tmp_list.append(value)
+            else:
+                # So far, only equals is supported by SMAC
+                legal = ['=', '>', '<']
```
mfeurer: The two lines above contradict each other. According to the format definition I got from Marius, > and < aren't part of this new pcs format.

ConfigSpace/io/pcs_new.py
```diff
@@ -0,0 +1,475 @@
...
+                    raise NotImplementedError()
+                tmp_list = []
+        configuration_space.add_forbidden_clause(ForbiddenAndConjunction(
+            *clause_list))
+
+
+    conditions_per_child = []
+    for condition in conditions:
+        conditions_per_child.append(condition)
+    for condition in conditions_per_child:
```
mfeurer: If I read the code below correctly, it only support two conditions. Could you please extend the code such that it supports an arbitrary amount of conditions?

ConfigSpace/io/pcs_new.py
```diff
@@ -0,0 +1,475 @@
...
+            param_lines.write("\n")
+
+    # Check if the default configuration is a valid configuration!
+
+
+    param_lines.seek(0)
+    return param_lines.getvalue()
+
+
+if __name__ == "__main__":
```
mfeurer: This code is not necessary any more.

ConfigSpace/hyperparameters.py
```diff
@@ -722,4 +724,167 @@ def get_neighbors(self, value, rs, number=np.inf, transform=False):
...
+        first element in our sequence as default.
+        """
+        if default is None:
+            return self.sequence[0]
+        elif self.is_legal(default):
+            return default
+        else:
+            raise ValueError("Illegal default value %s" % str(default))
+            
+    def get_seq_order(self):
```
mfeurer: Is this method used anywhere?

ConfigSpace/hyperparameters.py
```diff
@@ -722,4 +724,167 @@ def get_neighbors(self, value, rs, number=np.inf, transform=False):
...
+        """
+        return np.arange(1,self._num_elements+1)
+        
+    def get_order(self, value):
+        """
+        returns the seuence position/order of a certain value from the sequence
+        """
+        return self.value_dict[value]
+        
+    def get_value(self, idx):
```
mfeurer: Is this method used anywhere?

ConfigSpace/hyperparameters.py
```diff
@@ -722,4 +724,167 @@ def get_neighbors(self, value, rs, number=np.inf, transform=False):
...
+        """
+        returns the sequence value of a given order/position
+        """
+        return list(self.value_dict.keys())[list(self.value_dict.values()).index(idx)]
+            
+    def check_order(self,val1, val2):
```
mfeurer: Is this method used anywhere?

mfeurer:
Oh, and I just realized that the tests are failing. Could you please fix them as well?

mfeurer:
@Krxsy please ping me when I should review this PR again.

Krxsy:
@mfeurer  Yes, I will. I'm almost done (I'm tackling the condition extension right now)

Krxsy:
@mfeurer Hey, I think it should be ok now.

ConfigSpace/hyperparameters.py
```diff
@@ -722,4 +724,151 @@ def get_neighbors(self, value, rs, number=np.inf, transform=False):
...
+        repr_str.write(", Default: ")
+        repr_str.write(str(self.default))
+        repr_str.seek(0)
+        return repr_str.getvalue()
+
+    def is_legal(self, value):
+        """
+        checks if a certain value is represented in the sequence
+        """
+        if value in self.sequence:
```
mfeurer: You could do `return value in self.sequence`.

ConfigSpace/io/pcs_new.py
```diff
@@ -0,0 +1,452 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+__authors__ = ["Katharina Eggensperger", "Matthias Feurer", "Christina Hernandez Wunsch"]
```
mfeurer: If you change the encoding to UTF-8, you can write your name with `á`.

ConfigSpace/io/pcs_new.py
```diff
@@ -0,0 +1,452 @@
...
+                    j = i+1
+                else:
+                    i += 1
+        else:
+            conditions_per_child[child_name].append(condition)
+    for child_name in conditions_per_child:
+        condition_objects = []
+        for condition in conditions_per_child[child_name]:
+            child = configuration_space.get_hyperparameter(child_name)
+            if '(' not in condition:
```
mfeurer: Sorry, I don't understand why this can happen. Could you please add comments to this for-loop?

ConfigSpace/io/pcs_new.py
```diff
@@ -0,0 +1,452 @@
...
+                    elif operation == '<':
+                        condition = LessThanCondition(child, parent, restrictions)
+                    elif operation == '>':
+                        condition = GreaterThanCondition(child, parent, restrictions)
+        
+                    condition_objects.append(condition)       
+        if len(condition_objects) > 1:
+            for connective in connectives:
+                if connective == '&&':
+                    and_conjunction = AndConjunction(*condition_objects)
```
mfeurer: I don't think this is correct. This statement adds all conditions with the same boolean operator to the configuration space. Instead of doing this, you need to first add all adjacent `||` to an `OrConjunction` and then connect all left `&&` with a single `AndConjunction`.

ConfigSpace/io/pcs_new.py
```diff
@@ -0,0 +1,452 @@
...
+
+        # First build params
+        if param_lines.tell() > 0:
+            param_lines.write("\n")
+        if isinstance(hyperparameter, NumericalHyperparameter):
+            param_lines.write(build_continuous(hyperparameter))
+        elif isinstance(hyperparameter, CategoricalHyperparameter):
+            param_lines.write(build_categorical(hyperparameter))
+        elif isinstance(hyperparameter, Constant):
+            param_lines.write(build_constant(hyperparameter))
```
mfeurer: It looks like you've deleted the function, but still try to call it.

ConfigSpace/hyperparameters.py
```diff
@@ -722,4 +724,151 @@ def get_neighbors(self, value, rs, number=np.inf, transform=False):
...
+        return rs.randint(0, self._num_elements, size=size)
+
+    def has_neighbors(self):
+        """
+        checks if there are neighbors or we're only dealing with an 
+        one-element sequence
+        """
+        return len(self.sequence) > 1
+
+    def get_num_neighbors(self):
```
mfeurer: This method seems to be wrong. Assume that you have the following sequence: `[cold, luke-warm, warm]`. The length of the sequence is three, but `cold` only has one neighbour, namely `luke-warm`. Could you please write a test for this case and update this method?

ConfigSpace/io/pcs_new.py
mlindauer: Hi, as far as I understood the code, you allow for nested conditions (using brackets), rights? That is explicitly not supported by the new pcs format. We assume a CNF (conjunctive normal format) without a need for brackets for conditionals. This should make it a lot easier to parse and to evaluate conditions later --- essentially no need for recursion.

Krxsy: @mlindauer  @mfeurer Yes, exactly. Sorry, I will change that.( @sfalkner  )

sfalkner: Sorry Christina, I didn't know that SMAC 3 introduced yet another pcs format ... I, for once, liked the expressiveness of the SMAC 2.10 pcs format, but so be it ...

mlindauer: @sfalkner : the conditional format should be the same. also the smac 2.10 states: "There is no support for parenthesis with conditionals."

Krxsy:
@mfeurer I hope it's ok now

ConfigSpace/hyperparameters.py
```diff
@@ -836,11 +833,11 @@ def has_neighbors(self):
         """
         return len(self.sequence) > 1
 
-    def get_num_neighbors(self):
+    def get_num_neighbors(self, value):
```
mfeurer: I'm afraid this won't work, as the base class does not have this parameters. The only usage of this method is in the method [get_one_exchange_neighborhood](https://github.com/Krxsy/ConfigSpace/blob/d7663059cbec3cd4494c875f2773afe7059935c4/ConfigSpace/util.py#L96). Since it's possible to obtain the hyperparameter value in that function, could you please update that function and add an additional argument to each hyperparameter?

ConfigSpace/hyperparameters.py
mfeurer: Sorry, my bad. It is actually used.

ConfigSpace/hyperparameters.py
mfeurer: Okay, this one as well.

ConfigSpace/io/pcs_new.py
```diff
@@ -347,16 +354,40 @@ def read(pcs_string, debug=False):
...
-            for connective in connectives:
-                if connective == '&&':
-                    and_conjunction = AndConjunction(*condition_objects)
-                    configuration_space.add_condition(and_conjunction)
-                elif connective == '||':
+            if '||' in connection:
+                ands = []
+                ors = []
+                if '&&' in connection:
+                    if (len(connection) % 2 != 0) and (connection.count('&&') % 2 == 0):
```
mfeurer: When does this condition happen?

ConfigSpace/io/pcs_new.py
```diff
@@ -347,16 +354,40 @@ def read(pcs_string, debug=False):
...
+                                ands.append(condition_objects[i-1])
+                                ands.append(condition_objects[i])
+                                i += 2
+                                j += 1
+                            else:
+                                ors.append(condition_objects[i-1])
+                                i += 1
+                                j += 1
+
+                    orand_conjunction = OrConjunction(AndConjunction(*ands), *ors)
```
mfeurer: This only creates a single and conjunction inside an or conjunction. Instead, you want to connect multiple and conjunctions inside an or conjunction. Also, I think the logic of this code can be simplified. How about:
```
ands = []
ors = []
for i, conn in enumerate(connections):
    if conn == '||':
        ands.append(condition_objects[i])
        if len(ands) > 1:
            and = AndConjunction(*ands)
        else:
            and = ands[0]
        ors.append(and)
        ands = []
# Add the final object as well
ands.append(condition_objects[-1]
if len(ands) > 0:
    if len(ands) > 1:
        and = AndConjunction(*ands)
    else:
        and = ands[0]
    ors.append(and)
if len(ors) > 1:
    final_conjunction = OrConjunction(*ors)
else:
    final_conjunction = ors[0]
```
Also, I that there is no unit test for this piece of code. It would be good if you first write a few unit tests for several conditions that happen in practive.
```

**Comment by Krxsy** [comment]
@mfeurer I committed a simplified, more reasonable version. Are two test cases enough or should I add more?

File: ConfigSpace/io/pcs_new.py [hunk 27]
```diff
@@ -0,0 +1,443 @@
...
+            if '||' in str(condition):
+                ors = []
+                # 1st case we have a mixture of || and &&
+                if '&&' in str(condition):
+                    ors_combis = []
+                    for cond_parts in str(condition).split('||'):
+                        condition = str(cond_parts).split('&&')
+                        # if length is 1 it must be or
+                        if len(condition) == 1:
+                            element_list = [element for part in condition for element in part.split()]
```
mfeurer: This is hard for me to understand. Could you please add a comment on what this list comprehension looks like and what it is good for?

Krxsy: It converts the string which results from both splits (|| and &&) into a list of strings in order to process the condition type better. e.g. condition: ['classifier == k_nearest_neighbors '] ;  element_list = ['classifier', '==', 'k_nearest_neighbors']

sfalkner: Just to be nitpicky: if len(condition) equals 1 then the 'for part in condition' does nothing and the line could (probably) be just element_list = condition[0].split()
@Krxsy, you are thinking to complicated again :)

Krxsy: @sfalkner You're absolutely right.

Krxsy:
@mfeurer

mfeurer:
This looks good to me now. Thanks for the effort @Krxsy. I'll wait until I got coveralls working on this project until I merge this to make sure that the reader is actually well-tested.

mfeurer:
@Krxsy I don't understand why, but travis-ci doesn't run coveralls. Could you please merge the latest changes from the master branch into this branch?

Krxsy:
@mfeurer Now there seems to be an error in test_configuration_space.py ...

mfeurer:
That test failure seems to be unrelated to your changes, I'll take care of that when I'm back in Germany.

mfeurer:
@mlindauer I think you can try this now.

mlindauer:
@mfeurer Ok, I will do. I added an respective branch in the SMAC3 repo.

PR #9, opened 2017-01-13
Typecheck

smohsinali:


smohsinali:
. should I do something about travis path conflict because I changed it on my repository
. I use following command to check with mypy "mypy --silent-imports --check-untyped-defs util.py
" 
. in util.py mypy gives error at lines 127, 129, 130.

mfeurer:
* Please don't put any annotations in this PR, do separate PRs for the annotations
* Please don't change the paths in the readme etc.
* this does drop python2.7 from travis-ci, but does not remove the package six; please do this first.

PR #15, opened 2017-01-13
Python3

smohsinali:


mfeurer:
You're using `io.reduce` instead of `reduce`. Could you please change that?

smohsinali:
changed io.reduce to reduce

PR #16, opened 2017-01-13
Typecheck forbidden p3

smohsinali:


ConfigSpace/forbidden.py
```diff
@@ -29,14 +29,16 @@
...
 
-from ConfigSpace.hyperparameters import Hyperparameter
 
+from ConfigSpace.hyperparameters import Hyperparameter
+from typing import List, Dict, Any, Union
 
 class AbstractForbiddenComponent(object):
     __metaclass__ = ABCMeta
-
+    # todo : should absract methods be annotated with types?
```
mfeurer: From my point of view, this is not necessary here. Did mypy complain? If yes, we have to do something about it. Else, please remove the comment. The other abstract methods which should return something should be annotated though.

ConfigSpace/forbidden.py
```diff
@@ -72,12 +74,13 @@ def is_forbidden(self, instantiated_hyperparameters):
 
 
 class AbstractForbiddenClause(AbstractForbiddenComponent):
-    def get_descendant_literal_clauses(self):
+    def get_descendant_literal_clauses(self) -> List['AbstractForbiddenClause']:
```
mfeurer: Wouldn't `AbstractForbiddenComponent` be more appropriate her? Also, why is this a string? Moreover, the definition further below suggests that this function return `List[Hyperparameter].`

ConfigSpace/forbidden.py
```diff
@@ -72,12 +74,13 @@ def is_forbidden(self, instantiated_hyperparameters):
...
         return [self]
 
 
 class SingleValueForbiddenClause(AbstractForbiddenClause):
-    def __init__(self, hyperparameter, value):
+    # todo: possible types of hyperparameter valuees?
```
mfeurer: I'm afraid we need to accept `Any` here. Please remove the comment.

ConfigSpace/forbidden.py
```diff
@@ -89,7 +92,7 @@ def __init__(self, hyperparameter, value):
                              "'%s'" % (hyperparameter, str(value)))
         self.value = value
 
-    def is_forbidden(self, instantiated_hyperparameters, strict=True):
+    def is_forbidden(self, instantiated_hyperparameters: Hyperparameter, strict: bool=True) -> bool:
```
mfeurer: Please check the code by running it, but I think that `instantiated_hyperparameters` is a `dict`. Please also annotate `_is_forbidden`.

ConfigSpace/forbidden.py
```diff
@@ -109,7 +112,8 @@ def _is_forbidden(self, target_instantiated_hyperparameter):
 
 
 class MultipleValueForbiddenClause(AbstractForbiddenClause):
-    def __init__(self, hyperparameter, values):
+    # todo: type of vals of hp
```
mfeurer: Same as above.

ConfigSpace/forbidden.py
```diff
@@ -109,7 +112,8 @@ def _is_forbidden(self, target_instantiated_hyperparameter):
...
+    def __init__(self, hyperparameter: Hyperparameter, values: Any) -> None:
```
mfeurer: Same as above.

ConfigSpace/forbidden.py
```diff
@@ -182,7 +186,7 @@ def __init__(self, *args):
     def __repr__(self):
         pass
 
-    def get_descendant_literal_clauses(self):
+    def get_descendant_literal_clauses(self) -> List[Hyperparameter]:
```
mfeurer: Please update this based on the comment above.

ConfigSpace/forbidden.py
```diff
@@ -191,7 +195,7 @@ def get_descendant_literal_clauses(self):
                 children.append(component)
         return children
 
-    def is_forbidden(self, instantiated_hyperparameters, strict=True):
+    def is_forbidden(self, instantiated_hyperparameters: Hyperparameter, strict: bool=True) -> bool:
```
mfeurer: Same as above.

ConfigSpace/forbidden.py
```diff
@@ -210,8 +214,7 @@ def is_forbidden(self, instantiated_hyperparameters, strict=True):
         # outcomes
         evaluations = []
         for component in self.components:
-            e = component.is_forbidden(instantiated_hyperparameters,
-                                       strict=strict)
+            e = component.is_forbidden(instantiated_hyperparameters)
```
mfeurer: Why this?

ConfigSpace/forbidden.py
smohsinali: no mypy didnt complain...removed this todo

ConfigSpace/forbidden.py
smohsinali: changed. 
also its different class so no need to declare it as str.
why string? http://stackoverflow.com/questions/33533148/how-do-i-specify-that-the-return-type-of-a-method-is-the-same-as-the-class-itsel

ConfigSpace/forbidden.py
smohsinali: removed

ConfigSpace/forbidden.py
smohsinali: changed the type of instantiated_hyperparameters to Dict[str, Union[None, str, float, int]] .
all versions of _is_forbidden are annotated except abstract methods

ConfigSpace/forbidden.py
smohsinali: done

ConfigSpace/forbidden.py
smohsinali: done

ConfigSpace/forbidden.py
smohsinali: AbstractForbiddenComponent or AbstractForbiddenConjunction?

ConfigSpace/forbidden.py
smohsinali: type of strict is bool so its value should be 'true' or 'false' and not 'strict'

ConfigSpace/forbidden.py
mfeurer: Could you please check with the debugger and the unittests whether this actually returns a list of `AbstractForbiddenClause`? I would assume that the list can contain `AbstractForbiddenComponent` as well.

ConfigSpace/forbidden.py
```diff
@@ -182,8 +182,8 @@ def __init__(self, *args):
     @abstractmethod
     def __repr__(self):
         pass
-
-    def get_descendant_literal_clauses(self):
+    # todo: recheck is return type should be AbstractForbiddenComponent or AbstractForbiddenConjunction or Hyperparameter
```
mfeurer: Could you please check this with the debugger?

ConfigSpace/forbidden.py
mfeurer: Yes, but strict is a boolean variable that is passed to `is_forbidden` in line 195; so it won't change till down here.

smohsinali: fixed

ConfigSpace/forbidden.py
smohsinali: mypy accepts only List[hyperparameter] but according to debugger AbstractForbiddenComponent should also work

mfeurer: What do you mean by also? Can both happen?

smohsinali: from output of debugger it seems that the object of type forbidden has hyperparameter as its keys

smohsinali: fixed by adding hyperparameter attribute to AbstractForbiddenComponent

PR #20, opened 2017-01-13
Typecheck util p3

smohsinali:


ConfigSpace/util.py
```diff
@@ -122,12 +122,12 @@ def get_one_exchange_neighbourhood(configuration, seed):
                 # Activate hyperparameters if their parent node got activated
                 children = configuration.configuration_space.get_children_of(
                     hp_name)
-
+                # todo: debug why mypy is giving error here
                 if len(children) > 0:
-                    to_visit = deque()
+                    to_visit = deque()  # type : ignore
```
mfeurer: What happens if you set this to `type: deque`?

smohsinali: the problem was space between type and : 
also changed the type to deque

smohsinali: there was conflict in hyperparameter.py file...I resolved it so now it should merge

smohsinali:
there was conflict in hyperparameter.py file...I resolved it so now it should merge

PR #22, opened 2017-01-19
Irace p3

smohsinali:


mfeurer:
Before I'll review this, could you please add a few unit tests to make sure that this is working.

PR #23, opened 2017-02-03
use yield in get_one_exchange_neighbourhood

smohsinali:


test/converters_and_test_searchspaces/test_sample_configuration_spaces.py
```diff
@@ -57,9 +57,17 @@ def run_test(self):
                 c.is_valid_configuration()
                 neighborhood = ConfigSpace.util.get_one_exchange_neighbourhood(
                     c, seed=i)
-                np.random.shuffle(neighborhood)
-                for n in neighborhood[:5]:
+                # np.random.shuffle(neighborhood)
```
mfeurer: Please remove the comments and use the enumerate function to make this code shorter/easier.

ConfigSpace/util.py
```diff
@@ -81,137 +81,157 @@ def get_one_exchange_neighbourhood(configuration: Configuration, seed: int) -> L
...
+    # neighbourhood = []
+    # for i, hp_name in enumerate(configuration):
+    for i in range(hyperparameters_list_length):
+        index = np.random.randint(hyperparameters_list_length)
+        hp_name = hyperparameters_list[index]
+        if hp_name in neighbors_to_return:
+            np.random.shuffle(neighbors_to_return[hp_name])
+            n_ = neighbors_to_return[hp_name].pop()
+            if len(neighbors_to_return[hp_name]) == 0:
+                del neighbors_to_return[hp_name]
```
mfeurer: Don't delete here. If the list is empty, then you need to sample a different hyperparameter.

ConfigSpace/util.py
```diff
@@ -81,137 +81,157 @@ def get_one_exchange_neighbourhood(configuration: Configuration, seed: int) -> L
...
-        if not np.isfinite(array[i]):
-            continue
+    hyperparameters_list = list(configuration.keys())
+    hyperparameters_list_length = len(hyperparameters_list)
+    neighbors_to_return = dict()
+
+    # neighbourhood = []
+    # for i, hp_name in enumerate(configuration):
+    for i in range(hyperparameters_list_length):
+        index = np.random.randint(hyperparameters_list_length)
```
mfeurer: Don't use np.random.randint here, but use the random state from line 83.

ConfigSpace/util.py
```diff
@@ -81,137 +81,157 @@ def get_one_exchange_neighbourhood(configuration: Configuration, seed: int) -> L
...
+        hp_name = hyperparameters_list[index]
+        if hp_name in neighbors_to_return:
+            np.random.shuffle(neighbors_to_return[hp_name])
```
mfeurer: Same issue with the random state as above.

ConfigSpace/util.py
```diff
@@ -81,137 +81,157 @@ def get_one_exchange_neighbourhood(configuration: Configuration, seed: int) -> L
...
+                del neighbors_to_return[hp_name]
+            yield n_
+
+        else:
+            neighbourhood = []
+            number_of_sampled_neighbors = 0
+            array = configuration.get_array()
+
+            if not np.isfinite(array[index]):
+                continue
```
mfeurer: To make the code easier to comprehend, it would be good to add an empty list tho the neighbors_to_return dictionary.

ConfigSpace/util.py
```diff
@@ -81,137 +81,157 @@ def get_one_exchange_neighbourhood(configuration: Configuration, seed: int) -> L
...
+                    # Count iterations to not run into an infinite loop when
+                    # sampling floats/ints and there is large amount of forbidden
+                    #  values; also to find out if we tried to get a neighbor for
+                    #  a categorical hyperparameter, and the only possible
+                    # neighbor is forbidden together with another active
+                    # value/default hyperparameter
+                    iteration += 1
+            if len(neighbourhood) > 0:
+                neighbors_to_return[hp_name] = neighbourhood
+                np.random.shuffle(neighbors_to_return[hp_name])
```
mfeurer: Same issue with the random state here. Also, if you shuffle here, you don't need to shuffle in the code above.

ConfigSpace/util.py
```diff
@@ -81,137 +81,157 @@ def get_one_exchange_neighbourhood(configuration: Configuration, seed: int) -> L
...
+                n_ = neighbors_to_return[hp_name].pop()
+                if len(neighbors_to_return[hp_name]) == 0:
```
mfeurer: Same as above, you shouldn't delete this here but instead keep an empty array.

ConfigSpace/util.py
smohsinali: tried but it breaks the code

smohsinali:
PR is ready

PR #30, opened 2017-04-25
Vector representation checking

jtuyls:
Adds vector representation evaluation in conditions. Using this, sample_configuration can be made 20% faster. For now I added a new separate method sample_configuration_vector_checking so you can compare the performance of this method with the standard sample_configuration.

### DIFF
--- a/ConfigSpace/conditions.py
+++ b/ConfigSpace/conditions.py
@@ -114,7 +114,7 @@ class AbstractCondition(ConditionComponent):
         self.child_vector_id = None
         self.parent_vector_id = None
 
-    def set_vector_idx(self, hyperparameter_to_idx):
+    def set_vector_idx(self, hyperparameter_to_idx: dict):
         self.child_vector_id = hyperparameter_to_idx[self.child.name]
         self.parent_vector_id = hyperparameter_to_idx[self.parent.name]
 
@@ -175,7 +175,7 @@ class AbstractConjunction(ConditionComponent):
                 children.append(component)
         return children
 
-    def set_vector_idx(self, hyperparameter_to_idx):
+    def set_vector_idx(self, hyperparameter_to_idx: dict):
         for component in self.components:
             component.set_vector_idx(hyperparameter_to_idx)
 
--- a/ConfigSpace/conditions.py
+++ b/ConfigSpace/conditions.py
@@ -98,6 +98,18 @@ class AbstractCondition(ConditionComponent):
                              "different hyperparameters.")
         self.child = child
         self.parent = parent
+        self.child_vector_id = None
+        self.parent_vector_id = None
+
+    def set_vector_idx(self, hyperparameter_to_idx):
+        self.child_vector_id = hyperparameter_to_idx[self.child.name]
+        self.parent_vector_id = hyperparameter_to_idx[self.parent.name]
+
+    def get_children_vector(self) -> List[int]:
+        return [self.child_vector_id]
+
+    def get_parents_vector(self) -> List[int]:
+        return [self.parent_vector_id]
 
     def get_children(self) -> List[Hyperparameter]:
         return [self.child]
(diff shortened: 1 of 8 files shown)


ConfigSpace/conditions.py
```diff
@@ -176,6 +209,25 @@ def evaluate(self, instantiated_hyperparameters: Hyperparameter) -> bool:
 
         return self._evaluate(evaluations)
 
+    def evaluate_vector(self, instantiated_vector: List[int]) -> bool:
```
mfeurer: This should also be of type np.ndarray, right?

ConfigSpace/conditions.py
```diff
@@ -176,6 +209,25 @@ def evaluate(self, instantiated_hyperparameters: Hyperparameter) -> bool:
...
+        # Then, check if all parents were passed
+        conditions = self.get_descendant_literal_conditions()
+        for condition in conditions:
+            if condition.parent_vector_id not in range(len(instantiated_vector)):
+                raise ValueError("Evaluate must be called with all "
```
mfeurer: This error message isn't correct any more, right?

ConfigSpace/hyperparameters.py
```diff
@@ -70,6 +70,10 @@ def __repr__(self):
     def is_legal(self, value):
         raise NotImplementedError()
 
+    @abstractmethod
+    def is_legal_vector(self, value):
```
mfeurer: Sorry, I don't immediately see what `is_legal_vector` is doing. Could you please add a docstring?

mfeurer:
BTW: the increase in runtime was incredibly small, no need to worry about that.

ConfigSpace/conditions.py
jtuyls: Do you want to keep this check, because the instantiated vector is always the full configuration vector? It can happen that the vector parent id is None if the condition was not called to set it's vector id on time (however, this should not happen in the current implementation), but that is easier to check.

mfeurer: If this check isn't necessary any more, please remove it. A check whether the parent vector ID isnan would be good (hopefully doesn't take too much runtime...)

ConfigSpace/conditions.py
```diff
@@ -176,6 +222,22 @@ def evaluate(self, instantiated_hyperparameters: Hyperparameter) -> bool:
 
         return self._evaluate(evaluations)
 
+    def evaluate_vector(self, instantiated_vector: np.ndarray) -> bool:
+        # Then, check if all parents were passed
+        conditions = self.get_descendant_literal_conditions()
+        for condition in conditions:
+            if condition.parent_vector_id is None:
```
mfeurer: Can it happen that a `np.NaN` comes in here?

ConfigSpace/conditions.py
```diff
@@ -47,6 +49,17 @@ def __init__(self) -> None:
     @abstractmethod
     def __repr__(self) -> str:
         pass
+    @abstractmethod
+    def set_vector_idx(self, hyperparameter_to_idx) -> None:
```
mfeurer: Could you please check that all the arguments to the functions have a type annotation as well?

test/test_hyperparameters.py
```diff
@@ -578,6 +614,14 @@ def test_ordinal_is_legal(self):
...
+        # Test is legal vector
+        self.assertTrue(f1.is_legal_vector(1.0))
+        self.assertTrue(f1.is_legal_vector(0.0))
+        self.assertTrue(f1.is_legal_vector(0))
+        self.assertTrue(f1.is_legal_vector(3))
+        self.assertFalse(f1.is_legal_vector(-0.1))
+        self.assertFalse(f1.is_legal_vector("Hahaha"))
         
     def test_ordinal_check_order(self):
         f1 = OrdinalHyperparameter("temp",
```
mfeurer: Ordinals turned out to be tricky when we added them Could you please add a unit test for them as well?

mfeurer:
I have just merged a PR which improves the sampling procedure by 30-45% for Auto-sklearn. Could you please rebase your changes on this (it'll only affect the file `configuration_space.py`? You basically only need to copy your new way of obtaining the parent values to the correct place in the new code.

ConfigSpace/conditions.py
jtuyls: No, because the vector id indicates at which position in the vector the parent is located. This cannot be np.NaN

test/test_hyperparameters.py
jtuyls: This is a test case for the ordinal hyperparameter.

Issue #25, opened 2017-02-21
Expected equality after pcs-write/read?


shukon:
I stumbled over the following and I wondered whether this is expected behaviour:
When I write a ConfigSpace-object to a file using the `pcs.write()` and load it again with `pcs.read()`, `cs == cs_reloaded` evaluates to `False`. From some debugging I get that the attributes `_idx_to_hyperparameter`, `_parents` and `_hyperparameter_idx` are responsible (possibly others too).
I am not very familiar with the ConfigSpace-class, but intuitively this feels like a bug rather than a feature.

mfeurer:
This is certainly not intended. Is it possible for you to track the error down further or provide a minimal working/failing example?

shukon:
So, a minimum example using the spear-params-mixed.pcs (https://github.com/automl/ConfigSpace/blob/master/test/test_searchspaces/spear-params-mixed.pcs):

```
>>> import ConfigSpace
>>> import ConfigSpace.io.pcs as pcs
>>> with open("spear-params-mixed.pcs", 'r') as f: cs_one = pcs.read(f.readlines())
... 
>>> with open("example.pcs", 'w') as f: f.write(pcs.write(cs_one))
... 
1753
>>> with open("example.pcs", 'r') as f: cs_two = pcs.read(f.readlines())
... 
>>> repr(cs_one) == repr(cs_two)
True
>>> cs_one == cs_two
False
```

From the equality test of the `ConfigSpace` (https://github.com/automl/ConfigSpace/blob/master/ConfigSpace/configuration_space.py#L641-L649), elements of `__dict__` are compared. Investigating this, the following attributes seem to be responsible (except for random, which is already ruled out from equality-test):

```
>>> for attr in cs_one.__dict__:
...     if not getattr(cs_one, attr) == getattr(cs_two, attr):
...         print(attr)
... 
_hyperparameters
_conditionsals
_hyperparameter_idx
_idx_to_hyperparameter
_children
_parents
random
```

mfeurer:
@shukon I just pushed a potential fix, could you please check it?

shukon:
@mfeurer Two attributes remain problematic:
```
>>> for attr in cs_one.__dict__:
...     if not getattr(cs_one, attr) == getattr(cs_two, attr):
...         print(attr)
... 
_children
_parents
random
```

For `_parents` this is solely attributed to the order:
```
>>> print(cs_one._parents == cs_two._parents)
False
[...]

PR #50, opened 2017-06-27
ADD ordinals and unittests for new pcs

KEggensperger:
This PR adds the option to write ordinal parameters and adds some more unittests to test the latest pcs format.

### DIFF
--- a/test/io/test_pcs_converter.py
+++ b/test/io/test_pcs_converter.py
@@ -339,7 +339,65 @@ class TestPCSConverter(unittest.TestCase):
         cs.add_forbidden_clause(fb)
         value = pcs_new.write(cs)
         self.assertIn(expected, value)
-    
+
+    def test_build_new_GreaterThanFloatCondition(self):
+        expected = "b integer [0, 10] [5]\n" \
+                   "a real [0.0, 1.0] [0.5]\n\n" \
+                   "a | b > 5"
+        cs = ConfigurationSpace()
+        a = UniformFloatHyperparameter("a", 0, 1, 0.5)
+        b = UniformIntegerHyperparameter("b", 0, 10, 5)
+        cs.add_hyperparameter(a)
+        cs.add_hyperparameter(b)
+        cond = GreaterThanCondition(a, b, 5)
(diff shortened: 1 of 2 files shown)


test/io/test_pcs_converter.py
```diff
@@ -339,7 +339,65 @@ def test_build_new_forbidden(self):
...
+        cs = ConfigurationSpace()
+        a = UniformFloatHyperparameter("a", 0, 1, 0.5)
+        b = UniformIntegerHyperparameter("b", 0, 10, 5)
+        cs.add_hyperparameter(a)
+        cs.add_hyperparameter(b)
+        cond = GreaterThanCondition(a, b, 5)
+        cs.add_condition(cond)
+
+        value = pcs_new.write(cs)
+        self.assertIn(expected, value)
```
mfeurer: I'm wondering why you're not checking for equality here (and in all the other checks as well).

test/io/test_pcs_converter.py
```diff
@@ -436,5 +494,5 @@ def test_read_write(self):
             fh.write(pcs_string)
         with open(name, 'r') as fh:
             pcs_new = pcs.read(fh)
-        assert (pcs_new, cs)
+        self.assertTrue(pcs_new, cs)
```
mfeurer: I agree that the old check did nothing, but I think the new one is broken, too. I think you rather want `self.assertEqual(pcs_new, cs)` as the current one only check that the pcs_new string is not empty.

mlindauer:
ping....

test/io/test_pcs_converter.py
KEggensperger: Agree

test/io/test_pcs_converter.py
KEggensperger: @mfeurer Do you have any further ideas:
Fixing that check made it fail. I do not know why.

Looking at the output of print ```pcs_new``` and ```cs``` I could not find a difference.

mfeurer:
The failing test is related to #25.

PR #54, opened 2017-07-07
added: setting sampled value manually

Krxsy:
added the possibility to overwrite a specific value of a parameter manually after sampling the configuration.

e.g:
conf = pcs.sample_configuration()
conf['param_name'] = 0

[ as requested by @sfalkner ]

### DIFF
--- a/ConfigSpace/configuration_space.py
+++ b/ConfigSpace/configuration_space.py
@@ -1100,7 +1100,15 @@ class Configuration(object):
             return self[item]
         except:
             return default
-
+            
+    def __setitem__(self, key, value):
+        param = self.configuration_space.get_hyperparameter(key)
+        if param.is_legal(value):
+            self._values[key] = value
+            self._vector[self.configuration_space._hyperparameter_idx[key]] = self.configuration_space.get_hyperparameter(key)._inverse_transform(self[key])
+        else:
+            raise ValueError("Illegal value %s for hyperparameter %s" %(str(value), key))
+        
     def __contains__(self, item: str) -> bool:
         self._populate_values()
         return item in self._values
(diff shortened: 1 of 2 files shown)


sfalkner:
Thanks Christina. Just for completeness, could you please add a unit test that creates a simple config space, samples a config and then overwrites the values. Make sure that you cover all the following cases:
- a successful operation
- failed because the variable didn't exists (a typo in the name for example)
- failed because it's a invalid configuration (I don't know if is enough to specify an illegal value or whether the config should violate a constraint?
Thanks again!

Krxsy:
@sfalkner done

mfeurer:
Thanks for implementing this. I just checked (according to the comment of @sfalkner) what happens if the new value changes the conditionality structure. This is not accounted for. The code needs to look something like this (+changes in the utils file):
```
        param = self.configuration_space.get_hyperparameter(key)
        if not param.is_legal(value):
            raise ValueError(
                "Illegal value '%s' for hyperparameter %s" % (str(value), key))
        self._values[key] = value
        idx = self.configuration_space.get_idx_by_hyperparameter_name(key)
        vector_value = param._inverse_transform(value)
        new_array = self.get_array()
        new_array[idx] = vector_value
        from ConfigSpace.util import check_neighbouring_config_vector
        new_array = check_neighbouring_config_vector(self.configuration_space,
                                                     new_array, param.name)
        self._vector = new_array
        self._values = dict()
        self._query_values = False
        self.configuration_space.check_configuration(self)
```
If you want to I can take this over is I know the hidden secrets of the ConfigSpace a little bit better ;)

Krxsy:
@mfeurer  Oh, shoot! I completely forgot about that one. Thank you.

PR #55, opened 2017-07-10
Add Configuration.__setitem__

mfeurer:
Supersedes #54.


mfeurer:
@sfalkner @Krxsy please check if my changes make the function behave like you expect it.

mfeurer:
Good catch. This is obviously wrong and I changed the unit test to account for that (+ I fixed the bug).

PR #57, opened 2017-07-13
FIX #56; Write Conjunctions in new pcs format

### LINKED ISSUE #56: Broken pcs_new.write()
Hi, unfortunately, the pcs_new.write() writes weird conditions, e.g.:

`(@1:counter-bump | @1:S:counterCond == 'yes' && @1:counter-bump | @1:0:restarts in {'F', 'D', 'L', 'x', '+'})`

pcs_new.read() also cannot read these conditions and complains.
I think correct would be:
`@1:counter-bump | @1:S:counterCond == 'yes' && @1:0:restarts in {'F', 'D', 'L', 'x', '+'}`

I hope it is easy to fix.

I attach both the original pcs file (clasp in the old pcs format) and the pcs file generated by pcs_new.write()

[pcs_old.txt](https://github.com/automl/ConfigSpace/files/1145039/pcs_old.txt)
[pcs_new.txt](https://github.com/automl/ConfigSpace/files/1145038/pcs_new.txt)

KEggensperger:
Modified pcs_new writer to deal with Conjunctions. Not sure whether there is a better solution for L166 where I manually split a condition in order to get all variables (plus conditions) the child depends on.

### DIFF
--- a/ConfigSpace/io/pcs_new.py
+++ b/ConfigSpace/io/pcs_new.py
@@ -40,7 +40,6 @@ from ConfigSpace.conditions import EqualsCondition, NotEqualsCondition,\
 from ConfigSpace.forbidden import ForbiddenEqualsClause, \
     ForbiddenAndConjunction, ForbiddenInClause, AbstractForbiddenComponent, MultipleValueForbiddenClause
 
-
 # Build pyparsing expressions for params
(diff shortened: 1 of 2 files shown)


mfeurer:
There probably is a better solution to line 166, but I don't think it's worth spending the time on this right now. Given that a hyperparameter name should not contain `|`, I don't see an issue here.

PR #58, opened 2017-08-21
Cython forbidden

smohsinali:


mfeurer:
Hey, I think you forgot to check in the `.pyx` file. Could you please add that and then remove the `.so` and `.html` file? They're by-products and should not be in the repo.

smohsinali:
Oh! Ok. I will remove .html and add .pyx file but .so file is the compiled form of .pyx which is  actually used whenever forbidden_cython is imported. If we don't include it then user will have to first manullay compile it as part of the setup. Should I still ommit it or keep it.

mfeurer:
Ideally, this is automatically done in the `setup.py`, you can check Auto-sklearn's setup file to see how this can be done.

PR #65, opened 2018-01-04
Remove argparse from install_requires

podshumok:
argparse is a system library in Python2.7+ and Python3.2+
So if you don't support earlier versions, you don't need it. More than that dependency can pull out old version of argparse from PyPI and install it hiding system module


mfeurer:
Thanks a lot!

PR #82, opened 2018-07-25
Remove Cython and numpy build dependencies and implement work around suggested in

adithyabsk:
https://github.com/cython/cython/blob/master/docs/src/reference/compilation.rst
https://stackoverflow.com/questions/19919905/how-to-bootstrap-numpy-installation-in-setup-py
numpy/numpy#2434
numpy/numpy#432

In order to allow ConfigSpace to be installed using requirements file.

This change and the suggested change in autosklearn will allow the project to be independently installed.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,8 +1,7 @@
 from setuptools import setup, find_packages
 from setuptools.extension import Extension
+from setuptools.command.build_ext import build_ext as _build_ext
 import os
-from Cython.Build import cythonize
-import numpy as np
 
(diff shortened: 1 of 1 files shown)


adithyabsk:
Upon further research it seems to fail in awkward edge cases and I will find a workaround for my current issue until [PEP518](https://www.python.org/dev/peps/pep-0518/) is widely used. It seems that the current solution that most large frameworks implement is much more cumbersome than it's worth.

Issue #83, opened 2018-07-27
additional meta-data


janvanrijn:
Would it be possible to add a dict-field to each hyperparameter, that can be optionally left None? In order to keep using ConfigSpace in my projects, I sometimes find the need to store relevant meta-data about a hyperparameter (e.g., OpenML input id) and there is nowhere to store it.

mfeurer:
Sure, please go ahead and create a pull request.

PR #84, opened 2018-07-27
added support for meta-data in hyperparameters and configspace

janvanrijn:
As discussed in #83


janvanrijn:
Done. Any additional requests?

Issue #80, opened 2018-06-18
Conditional Parameters


janvanrijn:
One function that I commonly use within the OpenML spheres is 'filter setups according to config space'. For this I utilize the ConfigSpace library. I download a list of setups, give it a config space object and only keep the setups that are within this config space. I created a (non-perfect) version of it yesterday in the openml-python-contrib library:
https://github.com/openml/openml-python-contrib/blob/master/openmlcontrib/setups/functions.py

One thing that I couldn't figure out in the documentation of ConfigSpace: how to take into account hyperparameter conditionals. i.e., the libsvm_svc degree is only relevant when the poly kernel is selected. This knowledge is usually available in the Config Space, I just don't know how to extract it.

janvanrijn:
Thanks for your quick reply. However, I don't think that this does what I am looking for. 

On OpenML, we store for all flows programmatically all values for all hyperparameters. This could for example be a legal flow:

```
{
'kernel': 'rbf',
'degree': 1,
'gamma': 0.125,
'C': 4
}
```

Although it is a RBF kernel, it would have gotten the degree value of 1 (I guess because this might be the default in sklearn).

In the ConfigSpace lib, I now use the `ConfigSpace.Configuration(config_space, name_values_dict)` function to check whether this is a legal configuration. Unfortunately, it does not recognize that the value of degree should not matter, so I want to remove this from the `name_values_dict`. The actual question is how to recognize hyperparameters whos values do not matter (here: degree), because they are actually dependend on a value that was not selected (here: kernel, because value != poly).

janvanrijn:
I dived a little bit deeper into the code, and this part is what I need:

```
    def _check_configuration_rigorous(self, configuration: 'Configuration',
                                      allow_inactive_with_values: bool = False) -> None:
        vector = configuration.get_array()

        for hp_name, hyperparameter in self._hyperparameters.items():
            hp_value = vector[self._hyperparameter_idx[hp_name]]

            if not np.isnan(hp_value) and not hyperparameter.is_legal_vector(hp_value):
                raise ValueError("Hyperparameter instantiation '%s' "
                                 "(type: %s) is illegal for hyperparameter %s" %
                                 (hp_value, str(type(hp_value)),
                                  hyperparameter))

            conditions = self._parent_conditions_of[hyperparameter.name]

            active = True
            for condition in conditions:

                parent_vector_idx = condition.get_parents_vector()

                # if one of the parents is None, the hyperparameter cannot be
[...]

PR #85, opened 2018-07-27
modularized procedure to get names of active hyperparameters

janvanrijn:
Addresses #80


janvanrijn:
@mfeurer what do you think?

Issue #87, opened 2018-08-12
TypeError: 'module' object is not callable


burkartM:
Using add_configuration_space with a configspace containing an AbstractConjunction results in a TypeError: 'module' object is not callable in:
https://github.com/automl/ConfigSpace/blob/58321ea76e3c34043a9250ba5ccda1f880447abc/ConfigSpace/conditions.pyx#L472-L473

copy is not callable and should be used similar to:
https://github.com/automl/ConfigSpace/blob/58321ea76e3c34043a9250ba5ccda1f880447abc/ConfigSpace/conditions.pyx#L197-L202

mfeurer:
Thank you for reporting this issue. Would you like to create a pull request containing a fix for this issue?

burkartM:
After fixing the error I encountered a new problem:
In add_configspace add_conditions is called which does apparently not work with Conjunctions, you will lose all conditions but one.

I created a PR #88 that should fix both issues.

PR #88, opened 2018-08-15
Fixes for AbstractConjunctions bugs

burkartM:
Fixes for error and bug mentioned in #87

### DIFF
--- a/ConfigSpace/configuration_space.py
+++ b/ConfigSpace/configuration_space.py
@@ -233,7 +233,7 @@ class ConfigurationSpace(object):
                 edges.extend(
                     [(dlc.parent, dlc.child) for dlc in dlcs])
                 values.extend([dlc.value for dlc in dlcs])
-                conditions_to_add.extend(dlcs)
+                conditions_to_add.extend([condition] * len(dlcs))
 
         for edge, condition in zip(edges, conditions_to_add):
             self._check_condition(edge[1], condition)
--- a/ConfigSpace/conditions.pyx
+++ b/ConfigSpace/conditions.pyx
@@ -470,7 +470,7 @@ cdef class AbstractConjunction(ConditionComponent):
         return NotImplemented
 
     def __copy__(self):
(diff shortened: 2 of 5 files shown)


ConfigSpace/configuration_space.py
```diff
@@ -233,7 +233,7 @@ def add_conditions(self, conditions: List[ConditionComponent]) -> List[Condition
                 edges.extend(
                     [(dlc.parent, dlc.child) for dlc in dlcs])
                 values.extend([dlc.value for dlc in dlcs])
-                conditions_to_add.extend(dlcs)
+                conditions_to_add.extend([condition] * len(dlcs))
```
mfeurer: Could you please explain this statement?

mfeurer: Okay, this seems to fix your statement about adding Conjunctions. Could you please provide a failing example for this?

burkartM:
I added a unit test  that should fail on the "old" version.
Explanation for:
https://github.com/automl/ConfigSpace/blob/58321ea76e3c34043a9250ba5ccda1f880447abc/ConfigSpace/configuration_space.py#L236
I compared add_condition (which works) with add_condtions.
The only difference I could find is that while in add_condition you add for every the actual condition as condition:
https://github.com/automl/ConfigSpace/blob/58321ea76e3c34043a9250ba5ccda1f880447abc/ConfigSpace/configuration_space.py#L203-L207
you add the dlc as condition in add_conditions:
https://github.com/automl/ConfigSpace/blob/58321ea76e3c34043a9250ba5ccda1f880447abc/ConfigSpace/configuration_space.py#L236
https://github.com/automl/ConfigSpace/blob/58321ea76e3c34043a9250ba5ccda1f880447abc/ConfigSpace/configuration_space.py#L241-L242
This is why [condition] * len(dlcs) should fix the issue, because the zip will now map each edge to the actual condition and not the according dlc.

mfeurer:
Thanks a lot!

PR #112, opened 2019-03-08
[MRG] Refactor to fixes warnings in tests

thomasjpfan:
This PR contains the following:

1. Fixes warnings when running tests.
2. Removes print statements when running tests.
3. Fixes some Flake8 warnings
4. Adds language_level for `cython`
5. Moves to tempdir for `test_irace_writer.py`
6. Uses `pyx` extension for files that are compiled with Cython.
7. Adds test for public variables, such as `name`, `default_value`, etc.
8. Update to use language_level=3


mfeurer:
Thanks a lot for your work, this looks great. I've got two questions:

1. Would there be any effect in using language level 3 instead of 2? (or differently, is there a reason you chose 2 over 3?)
2. Would you mind adding a flake8 test to enforce proper formatting of the code in the future (for example as done in the [OpenML-Python repository](https://github.com/openml/openml-python/))?

thomasjpfan:
1. Since the community is moving to Python 3, it is nice to update the cython language level to 3 as well. Functionally, changing the language level will cause python 2 only syntax to fail compilation.

2. I will add a flake8 test to enforce proper formatting. 

It looks like the line length is set to 120, is this correct?

mfeurer:
Thanks for the quick answer. Is there any Python 2 syntax? I would expect no? And regarding the limit, I don't think we ever enforced one. If 120 is natural given the current code style, please use that.

thomasjpfan:
This PR has been updated with:

1. Tests for public attributes.
2. Update cython files to language level 3.
3. Adds flake8 to travis ci.

mfeurer:
This look like a real improvement to the code base, thanks a lot! I was a bit puzzled by setup.py being able to compile the file util.py which was named util.pyx, could you please give me a brief explanation or post a link on why this works?

thomasjpfan:
A section of the `setup.py` file states which files need to be "cythonized":

https://github.com/automl/ConfigSpace/blob/76a6941d15ee84c2bfd93bfb56560afd7a2a298c/setup.py#L21-L40

Originally, they were targeting files such as `util.py` which allows it to be compiled. When this PR changed the names to `util.pyx`, the `setup.py` was updated with these new names.

mfeurer:
Got it, I didn't see that you renamed the files. Thanks a lot for the explanation.

PR #118, opened 2019-06-07
Fix error message regarding swapped upper/lower bounds in hyperparameters.pyx

scriptator:
Fixed an error message in `hyperparameters.pyx` where upper and lower bound in string formatting was swapped.


mfeurer:
Thanks for the fix. I added it by merging it locally.

Issue #119, opened 2019-06-27
Sampling problem with quantized UniformIntegerHyperparameter


d060030:
### Context
Configuration space with a uniform integer hyperparameter with quantization (see example code below)

### Problem
When sampling more than one configuration at a time (size>1), a type error is thrown (see stack trace below). The error seems to be related to the transformation and following the inverse transformation of the underlying sampled raw float value(s) and a mismatch of parameter types (vector vs. scalar). To my knowledge, only the UniformIntegerHyperparameter class is affected.

#### Example code
```
import ConfigSpace as CS
import ConfigSpace.hyperparameters as CSH

cs = CS.ConfigurationSpace()
cs.add_hyperparameter(CSH.UniformIntegerHyperparameter("a", lower=1, upper=10, q=1))

cs.sample_configuration(size=1) # works
cs.sample_configuration(size=2) # fails
```

#### Stack trace for example code
```
Traceback (most recent call last):
  File "main.py", line 8, in <module>
    cs.sample_configuration(size=2) # fails
  File "ConfigSpace/configuration_space.pyx", line 1265, in ConfigSpace.configuration_space.ConfigurationSpace.sample_configuration
  File "ConfigSpace/hyperparameters.pyx", line 799, in ConfigSpace.hyperparameters.UniformIntegerHyperparameter._sample
  File "ConfigSpace/hyperparameters.pyx", line 806, in ConfigSpace.hyperparameters.UniformIntegerHyperparameter._transform
  File "ConfigSpace/hyperparameters.pyx", line 495, in ConfigSpace.hyperparameters.UniformFloatHyperparameter._transform
TypeError: only size-1 arrays can be converted to Python scalars
```

mfeurer:
Thanks a lot for reporting and fixing this!

PR #120, opened 2019-06-27
Fix quantization in UniformIntegerHyperparameter sampling

d060030:
This pull request tries to fix an issue when sampling more than one configuration at a time (size>1), see issue #119.


mfeurer:
Thanks a lot for this fix!

PR #127, opened 2019-09-29
Fix huge slowdown in generating neighbors

Yatoom:
The unnecessary calls to `np.any()` caused a huge slowdown in generating neighbors in SMAC. See https://github.com/automl/SMAC3/issues/540. I added a check to see if this call is needed.


mfeurer:
Thanks a lot, this even makes the unit tests faster!

PR #129, opened 2019-10-01
More performance improvements for generating neighbors

Yatoom:
I have added more performance improvements and cleaned up the code for the `_transform()` functions so that it is more consistent and has less code repetition.

I ran the unit tests locally on my computer and found that running all cases took **~36 seconds** with the code in my previous pull request (#127). This is now further reduced to **~16 seconds**.

### DIFF
--- a/ConfigSpace/hyperparameters.pyx
+++ b/ConfigSpace/hyperparameters.pyx
@@ -377,7 +377,7 @@ cdef class IntegerHyperparameter(NumericalHyperparameter):
             return None
         return self._transform_scalar(vector)
 
-    cpdef int _transform_scalar(self, double scalar):
+    cpdef long _transform_scalar(self, double scalar):
         raise NotImplementedError()
 
     cpdef np.ndarray _transform_vector(self, np.ndarray vector):
@@ -854,7 +854,7 @@ cdef class UniformIntegerHyperparameter(IntegerHyperparameter):
 
         return np.rint(vector)
 
-    cpdef int _transform_scalar(self, double scalar):
+    cpdef long _transform_scalar(self, double scalar):
         if not math.isfinite(scalar):
             raise ValueError()
         scalar = self.ufhp._transform_scalar(scalar)
@@ -1167,7 +1167,7 @@ cdef class NormalIntegerHyperparameter(IntegerHyperparameter):
         vector = self.nfhp._transform_vector(vector)
         return np.rint(vector)
 
-    cpdef int _transform_scalar(self, double scalar):
+    cpdef long _transform_scalar(self, double scalar):
         if not math.isfinite(scalar):
             raise ValueError()
         scalar = self.nfhp._transform_scalar(scalar)
@@ -1605,7 +1605,7 @@ cdef class OrdinalHyperparameter(Hyperparameter):
                              'hyperparameter %s with an integer, but provided '
                              'the following float: %f' % (self, vector))
 
-    cpdef int _transform_scalar(self, double scalar):
+    cpdef long _transform_scalar(self, double scalar):
         if not math.isfinite(scalar):
             raise ValueError('number %s contains non-finite numbers' % scalar)
 
--- a/ConfigSpace/hyperparameters.pyx
+++ b/ConfigSpace/hyperparameters.pyx
@@ -508,7 +508,7 @@ cdef class UniformFloatHyperparameter(FloatHyperparameter):
 
     cpdef np.ndarray _transform_vector(self, np.ndarray vector):
         if np.isnan(vector).any():
-            raise ValueError()
+            raise ValueError('Vector %s contains NaN\'s' % vector)
         vector = vector * (self._upper - self._lower) + self._lower
         if self.log:
             vector = np.exp(vector)
@@ -517,8 +517,8 @@ cdef class UniformFloatHyperparameter(FloatHyperparameter):
         return np.maximum(self.lower, np.minimum(self.upper, vector))
 
     cpdef double _transform_scalar(self, double scalar):
-        if scalar == np.nan:
-            raise ValueError()
+        if scalar != scalar:
+            raise ValueError('Number %s is NaN' % scalar)
         scalar = scalar * (self._upper - self._lower) + self._lower
         if self.log:
             scalar = math.exp(scalar)
@@ -713,7 +713,7 @@ cdef class NormalFloatHyperparameter(FloatHyperparameter):
 
     cpdef np.ndarray _transform_vector(self, np.ndarray vector):
         if np.isnan(vector).any():
-            raise ValueError()
+            raise ValueError('Vector %s contains NaN\'s' % vector)
         if self.log:
             vector = np.exp(vector)
         if self.q is not None:
@@ -721,8 +721,8 @@ cdef class NormalFloatHyperparameter(FloatHyperparameter):
         return vector
 
     cpdef double _transform_scalar(self, double scalar):
-        if scalar == np.nan:
-            raise ValueError()
+        if scalar != scalar:
+            raise ValueError('Number %s is NaN' % scalar)
         if self.log:
             scalar = math.exp(scalar)
         if self.q is not None:
@@ -1347,7 +1347,7 @@ cdef class CategoricalHyperparameter(Hyperparameter):
 
     cpdef np.ndarray _transform_vector(self, np.ndarray vector):
         if np.isnan(vector).any():
-            raise ValueError('vector %s contains non-finite numbers' % vector)
+            raise ValueError('Vector %s contains NaN\'s' % vector)
 
         if np.equal(np.mod(vector, 1), 0):
             return self.choices[vector.astype(int)]
@@ -1357,8 +1357,8 @@ cdef class CategoricalHyperparameter(Hyperparameter):
                              'the following float: %f' % (self, vector))
 
     def _transform_scalar(self, scalar: Union[float, int]) -> Union[float, int, str]:
-        if scalar == np.nan:
-            raise ValueError('number %s contains non-finite numbers' % scalar)
+        if scalar != scalar:
+            raise ValueError('Number %s is NaN' % scalar)
 
         if scalar % 1 == 0:
             return self.choices[int(scalar)]
@@ -1587,7 +1587,7 @@ cdef class OrdinalHyperparameter(Hyperparameter):
 
     cpdef np.ndarray _transform_vector(self, np.ndarray vector):
         if np.isnan(vector).any():
-            raise ValueError('vector %s contains non-finite numbers' % vector)
+            raise ValueError('Vector %s contains NaN\'s' % vector)
 
         if np.equal(np.mod(vector, 1), 0):
             return self.sequence[vector.astype(int)]
@@ -1597,8 +1597,8 @@ cdef class OrdinalHyperparameter(Hyperparameter):
                              'the following float: %f' % (self, vector))
 
     def _transform_scalar(self, scalar: Union[float, int]) -> Union[float, int, str]:
-        if scalar == np.nan:
-            raise ValueError('number %s contains non-finite numbers' % scalar)
+        if scalar != scalar:
+            raise ValueError('Number %s is NaN' % scalar)
 
         if scalar % 1 == 0:
             return self.sequence[int(scalar)]
--- a/ConfigSpace/hyperparameters.pyx
+++ b/ConfigSpace/hyperparameters.pyx
@@ -27,7 +27,7 @@
 # SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 # cython: language_level=3
-
+import math
 from abc import ABCMeta, abstractmethod
 import warnings
 
@@ -95,6 +95,14 @@ cdef class Hyperparameter(object):
 
     def _transform(self, vector: Union[np.ndarray, float, int]) -> Optional[Union[np.ndarray, float, int]]:
         raise NotImplementedError()
+
+    def _check_vector(self, vector: Union[np.ndarray, float, int]) -> Optional[Union[np.ndarray, float, int]]:
+        is_array = isinstance(vector, np.ndarray)
+
+        if is_array:
+            return is_array, np.any(~np.isfinite(vector))
+
+        return is_array, not math.isfinite(vector)
 
     def _inverse_transform(self, vector):
         raise NotImplementedError()
@@ -197,7 +205,8 @@ cdef class Constant(Hyperparameter):
         return 0 if size == 1 else np.zeros((size,))
 
     def _transform(self, vector: Union[np.ndarray, float, int]) -> Optional[Union[np.ndarray, float, int]]:
-        if not np.isfinite(vector):
+        is_array, not_finite = self._check_vector(vector)
+        if not_finite:
             return None
         return self.value
 
@@ -319,6 +328,13 @@ cdef class NumericalHyperparameter(Hyperparameter):
             log=self.log,
             q=self.q
         )
+
+    def _apply_quantization_factor(self, vector: Union[np.ndarray, float, int], is_array: bool, q: Optional[int]) -> Union[np.ndarray, int]:
+        if q is not None:
+            if is_array:
+                return np.rint(vector / q) * q
+            return int(round(vector / q)) * q
+        return vector
 
 
 cdef class FloatHyperparameter(NumericalHyperparameter):
@@ -486,19 +502,21 @@ cdef class UniformFloatHyperparameter(FloatHyperparameter):
         return rs.uniform(size=size)
 
     def _transform(self, vector: Union[np.ndarray, float, int]) -> Optional[Union[np.ndarray, float, int]]:
-        if isinstance(vector, np.ndarray):
-            if np.any(np.isnan(vector)):
-                return None
-        elif np.isnan(vector):
+
+        is_array, not_finite = self._check_vector(vector)
+
+        if not_finite:
             return None
-        vector *= (self._upper - self._lower)
-        vector += self._lower
+
+        vector = vector * (self._upper - self._lower) + self._lower
+
         if self.log:
-            vector = np.exp(vector)
-        if self.q is not None:
-            vector = np.rint(vector / self.q) * self.q
-        vector = np.minimum(self.upper, vector)
-        vector = np.maximum(self.lower, vector)
+            vector = np.exp(vector) if is_array else math.exp(vector)
+
+        vector = self._apply_quantization_factor(vector, is_array, self.q)
+
+        vector = np.maximum(self.lower, np.minimum(self.upper, vector)) if is_array else \
+            max(self.lower, min(self.upper, vector))
         return vector
 
     def _inverse_transform(self, vector: Union[np.ndarray, None]) -> Union[np.ndarray, float, int]:
@@ -686,12 +704,16 @@ cdef class NormalFloatHyperparameter(FloatHyperparameter):
         return rs.normal(mu, sigma, size=size)
 
     def _transform(self, vector: Union[np.ndarray, float, int]) -> Optional[Union[np.ndarray, float, int]]:
-        if np.isnan(vector):
+
+        is_array, not_finite = self._check_vector(vector)
+
+        if not_finite:
             return None
+
         if self.log:
-            vector = np.exp(vector)
-        if self.q is not None:
-            vector = int(np.round(vector / self.q, 0)) * self.q
+            vector = np.exp(vector) if is_array else math.exp(vector)
+
+        vector = self._apply_quantization_factor(vector, is_array, self.q)
         return vector
 
     def _inverse_transform(self, vector: Optional[np.ndarray]) -> Union[float, np.ndarray]:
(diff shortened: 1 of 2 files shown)


mfeurer:
Thanks a lot for the patch. Unfortunately, I cannot confirm any speed improvements on my machine. I tried the unit tests in `test.test_converters_and_test_searchspaces` which are somewhat realistic and also the script `scripts.benchmark_sampling` and found that the tests go from 41s to 42s, while the benchmark sampling stays approximately the same. Could you please check these numbers, too, and also paste the output of the script `benchmark_sampling.py` with and without the change?

My outputs with this PR:
```
###
/home/feurerm/sync_dir/projects/ConfigSpace/test/test_searchspaces/auto-sklearn_2017_11_17.pcs
Average time sampling 100 configurations 0.01909217834472656
Average time retrieving a nearest neighbor 0.005693285465240478
Average time checking one configuration 0.00031762908805500376
```
without:
```
###
/home/feurerm/sync_dir/projects/ConfigSpace/test/test_searchspaces/auto-sklearn_2017_11_17.pcs
Average time sampling 100 configurations 0.018399810791015624
Average time retrieving a nearest neighbor 0.0057344818115234375
Average time checking one configuration 0.00029520517346834894
```

I have some further ideas on how to improve the performance of the transform functions that I'd be happy to share if you're interested, but they might take a bit more time to implement.

Yatoom:
Thanks, I will look into it soon! And I would be happy to hear your ideas.

mfeurer:
I think a lot of slowdown here comes from having a single method to handle both the array and the non-array case, which results in quite some overhead. 

Instead, if one would have a function `_transform_vector` and `_transform_scalar` one could save quite some time because one could also make the functions `cpdef` functions as there would only be a single possible output type. An alternative implementation could also use only arrays to not have two strains of logic.

Yatoom:
Hi @mfeurer, I investigated the `benchmark_sampling.py` results. It seems that the times are indeed the same. There are two explanations:

- The time it costs to execute the transform function is insignificant compared to the other code in `sample_configuration()`. Executing the transform function takes a factor of 10^-6 to 10^-7 seconds, while the average time of sampling 100 configurations takes 0.018 seconds, or 0.00018 ≈ 10^-4 seconds per configuration.
- 723d3cd6f4174cd742d829b87e6fb4cb1d03b299 already includes a performance improvement for Uniform Float Hyperparameters. Furthermore, the Uniform Integer Hyperparameters were already optimized and the Categorical Hyperparameters stayed the same.

However, looking at the `_transform()` function independently, I did see an improvement in speed. For my tests, I used the `_transform()` function of the Uniform Float Hyperparameter. With 723d3cd6f4174cd742d829b87e6fb4cb1d03b299, we get a speedup of a factor ~2.7. With the current pull request, we get a speedup of ~8.7 and with your idea we get a speed up of ~18.2 :smiley: 

I could split all the `_transform()` functions into `_transform_vector()` and `_transform_scalar()` and update the PR. And I could also try to figure out when and why the performance of the transform function affects the overall performance, as it does for example seem to speedup running all test-cases. What do you think?

mfeurer:
> With the current pull request, we get a speedup of ~8.7 and with your idea we get a speed up of ~18.2

That sounds great!

> Executing the transform function takes a factor of 10^-6 to 10^-7 seconds, while the average time of sampling 100 configurations takes 0.018 seconds, or 0.00018 ≈ 10^-4 seconds per configuration.

Keep in mind that this is also important for generating nearest neighbor (for SMAC), and I assume it has a larger effect there.

> And I could also try to figure out when and why the performance of the transform function affects the overall performance, as it does for example seem to speedup running all test-cases.

I think this would be most important.

> I could split all the _transform() functions into _transform_vector() and _transform_scalar() and update the PR.

From a maintenance point I would definitely be in favor of this refactoring as it will make the code clearer an replace a few if/else statements. One can then also start typing the transform functions.

mfeurer:
So I'd definitely appreciate an updated PR, but I think you need to decide if it's worth for you splitting up the other transform functions.

Yatoom:
Hi @mfeurer, I have split everything into scalar and vector transforms and debugged most problems I encountered, but I'm still failing one check: in `test_sample_UniformFloatHyperparameter` it seems that not all bins are filled. 

For example, the second `counts_per_bin` list looks like this: `[9245, 0, 9025, 0, 9300, 0, 8997, 0, 8991, 0, 9001, 0, 9113, 9056, 0, 0, 9118, 9052, 0, 0, 9102]`. 

Do you have an idea of what could be wrong here?

mfeurer:
Do the unit tests work locally for you? Currently, they are killed on travis-ci which makes it really hard to see what's going on. Could you please check?

Yatoom:
Hi @mfeurer, it turned out to be a problem with precision. Cython uses 32 bit floats & Python uses 64 bit floats, and this caused the Cython floats to be rounded down to the wrong bucket. So the solution was to use double's instead of float's. And I fixed some other problems as well.

ConfigSpace/hyperparameters.pyx
```diff
@@ -26,19 +26,16 @@
...
-from ConfigSpace.hyperparameters cimport Hyperparameter
-
 from collections import OrderedDict, Counter
-import copy
 from typing import List, Any, Dict, Union, Set, Tuple, Optional
-import io
-import numpy as np
+
 cimport numpy as np
+import numpy as np
```
mfeurer: Is there a reason to swap these around?

ConfigSpace/hyperparameters.pyx
```diff
@@ -326,18 +323,31 @@ cdef class FloatHyperparameter(NumericalHyperparameter):
...
     cpdef bint is_legal_vector(self, DTYPE_t value):
-        raise NotImplemented
+        raise NotImplementedError()
 
     def check_default(self, default_value: Union[int, float]) -> float:
-        raise NotImplemented
+        raise NotImplementedError()
 
+    def _transform(self, vector: Union[np.ndarray, float, int]) -> Optional[Union[np.ndarray, float, int]]:
+        if isinstance(vector, np.ndarray):
```
mfeurer: Is there a reason to not use `_check_vector` here?

ConfigSpace/hyperparameters.pyx
```diff
@@ -358,6 +368,21 @@ cdef class IntegerHyperparameter(NumericalHyperparameter):
                              " %s." % (name, type(parameter), str(parameter)))
         return int(parameter)
 
+    def _transform(self, vector: Union[np.ndarray, float, int]) -> Optional[Union[np.ndarray, float, int]]:
+        if isinstance(vector, np.ndarray):
```
mfeurer: Is there a reason to not use `_check_vector` here?

ConfigSpace/hyperparameters.pyx
```diff
@@ -485,21 +510,26 @@ cdef class UniformFloatHyperparameter(FloatHyperparameter):
...
-    def _transform(self, vector: Union[np.ndarray, float, int]) -> Optional[Union[np.ndarray, float, int]]:
-        if isinstance(vector, np.ndarray):
-            if np.any(np.isnan(vector)):
-                return None
-        elif np.isnan(vector):
-            return None
-        vector *= (self._upper - self._lower)
-        vector += self._lower
+    cpdef np.ndarray _transform_vector(self, np.ndarray vector):
+        if np.any(~np.isfinite(vector)):
```
mfeurer: Is this strictly necessary? This function is right now only called from `self._transform` which makes sure that there are no `np.NaN`s, right?

ConfigSpace/hyperparameters.pyx
```diff
@@ -685,15 +715,24 @@ cdef class NormalFloatHyperparameter(FloatHyperparameter):
         sigma = self.sigma
         return rs.normal(mu, sigma, size=size)
 
-    def _transform(self, vector: Union[np.ndarray, float, int]) -> Optional[Union[np.ndarray, float, int]]:
-        if np.isnan(vector):
-            return None
+    cpdef np.ndarray _transform_vector(self, np.ndarray vector):
+        if np.any(~np.isfinite(vector)):
```
mfeurer: Same as above.

ConfigSpace/hyperparameters.pyx
```diff
@@ -806,23 +845,22 @@ cdef class UniformIntegerHyperparameter(IntegerHyperparameter):
         value = self._inverse_transform(value)
         return value
 
-    def _transform(self, vector: Union[np.ndarray, float, int]) -> Optional[Union[np.ndarray, float, int]]:
+    cpdef np.ndarray _transform_vector(self, np.ndarray vector):
+        if np.any(~np.isfinite(vector)):
```
mfeurer: Same as above.

ConfigSpace/hyperparameters.pyx
Yatoom: The `_transform_vector()` and `_transform_scalar()` methods are now defined as cpdef's and check if the input is finite by themselves, i.e. accelerated by Cython. They raise a `ValueError` if the input is not finite. The `_transform()`  method here should instead catch the error and return None.

ConfigSpace/hyperparameters.pyx
Yatoom: I think this was just done by auto-formatting, so there's no special reason behind it.

ConfigSpace/hyperparameters.pyx
Yatoom: Same as above.

ConfigSpace/hyperparameters.pyx
Yatoom: Yeah, so that should be removed in the `_transform()` function, so we can check it inside the cpdef function. The idea is that the `transform_vector()` and `transform_scalar()` can also be called without calling `_transform()` first, i.e. when you can already guarantee yourself that the input is a scalar or a vector.

ConfigSpace/hyperparameters.pyx
Yatoom: Same as above.

ConfigSpace/hyperparameters.pyx
Yatoom: Same as above.

ConfigSpace/hyperparameters.pyx
mfeurer: > Furthermore, then you can call transform_vector() and transform_scalar() without calling _transform() first, i.e. when you can already guarantee yourself that the input is a scalar or a vector.

Yep, that'll be a great way to improve the speed of sampling and/or neighborhood generation. I'm really looking forward to that.

ConfigSpace/hyperparameters.pyx
```diff
@@ -806,23 +843,22 @@ cdef class UniformIntegerHyperparameter(IntegerHyperparameter):
...
```
mfeurer: Hm, I just did some checks and think this can be sped up:
```
In [6]: %timeit np.any(np.isfinite(a))                                                                   
4.58 µs ± 45.7 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [9]: %timeit (~np.isfinite(a)).any()                                                                  
3.14 µs ± 46.6 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [4]: %timeit np.isnan(a).any()                                                                        
2.52 µs ± 21.3 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

mfeurer: Or like this:
```
In [11]: %timeit not np.isfinite(a).all()                                                             
2.73 µs ± 141 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

ConfigSpace/hyperparameters.pyx
```diff
@@ -806,23 +843,22 @@ cdef class UniformIntegerHyperparameter(IntegerHyperparameter):
...
-        if isinstance(vector, np.ndarray):
-            if np.any(np.isnan(vector)):
-                return None
-        elif np.isnan(vector):
-            return None
+        return np.rint(vector)
 
-        vector = self.ufhp._transform(vector)
+    cpdef long _transform_scalar(self, double scalar):
+        if not math.isfinite(scalar):
```
mfeurer: I'm only seeing this now, but you replaced the check for NaN with a check for being finite. I don't think this is necessary as there are usually no infs/-infs around. Also, it's much faster to only check for NaN:
```
In [13]: %timeit math.isfinite(np.NaN)                                                                
127 ns ± 1.93 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [14]: %timeit np.NaN == np.NaN                                                                     
89.8 ns ± 0.833 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```

ConfigSpace/hyperparameters.pyx
```diff
@@ -1123,20 +1159,17 @@ cdef class NormalIntegerHyperparameter(IntegerHyperparameter):
         value = self._inverse_transform(value)
         return value
 
-    def _transform(self, vector: Union[np.ndarray, float, int]) -> Optional[Union[np.ndarray, int]]:
+    cpdef np.ndarray _transform_vector(self, np.ndarray vector):
+        if np.any(~np.isfinite(vector)):
```
mfeurer: Is this check necessary? It's also conducted in the `_transform_vector` method of `nfhp`, right?

ConfigSpace/hyperparameters.pyx
```diff
@@ -1123,20 +1159,17 @@ cdef class NormalIntegerHyperparameter(IntegerHyperparameter):
...
-        elif np.isnan(vector):
-            return None
-
-        vector = self.nfhp._transform(vector)
-        vector = (np.round(vector, 0)).astype(int)
-        if isinstance(vector, (np.int, np.int32, np.int64)):
-            vector = int(vector)
-        return vector
+    cpdef long _transform_scalar(self, double scalar):
+        if not math.isfinite(scalar):
```
mfeurer: Same as above.

ConfigSpace/hyperparameters.pyx
```diff
@@ -1322,16 +1355,36 @@ cdef class CategoricalHyperparameter(Hyperparameter):
...
-            return None
+    cpdef np.ndarray _transform_vector(self, np.ndarray vector):
+        if np.any(~np.isfinite(vector)):
+            raise ValueError('vector %s contains non-finite numbers' % vector)
+
         if np.equal(np.mod(vector, 1), 0):
-            return self.choices[int(vector)]
-        else:
-            raise ValueError('Can only index the choices of the categorical '
+            return self.choices(vector.astype(int))
```
mfeurer: Why did you replace the way the choices array is accessed (via `[]`) by `()`?

ConfigSpace/hyperparameters.pyx
```diff
@@ -1542,16 +1595,36 @@ cdef class OrdinalHyperparameter(Hyperparameter):
...
-            return None
+    cpdef np.ndarray _transform_vector(self, np.ndarray vector):
+        if np.any(~np.isfinite(vector)):
+            raise ValueError('vector %s contains non-finite numbers' % vector)
+
         if np.equal(np.mod(vector, 1), 0):
-            return self.sequence[int(vector)]
-        else:
-            raise ValueError('Can only index the choices of the ordinal '
+            return self.sequence(vector.astype(int))
```
mfeurer: Same as above.

ConfigSpace/hyperparameters.pyx
```diff
@@ -197,7 +199,8 @@ cdef class Constant(Hyperparameter):
         return 0 if size == 1 else np.zeros((size,))
 
     def _transform(self, vector: Union[np.ndarray, float, int]) -> Optional[Union[np.ndarray, float, int]]:
-        if not np.isfinite(vector):
+        is_array, not_finite = self._check_vector(vector)
```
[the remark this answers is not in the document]
Yatoom: If we split, then both `transform_vector` and `transform_scalar` have to return `self.value` in the `Constant` class, which can be of multiple types right? Then these methods should both be normal Python methods. It still makes sense to make this split for compatibility though.

ConfigSpace/hyperparameters.pyx
```diff
@@ -96,6 +91,14 @@ cdef class Hyperparameter(object):
     def _transform(self, vector: Union[np.ndarray, float, int]) -> Optional[Union[np.ndarray, float, int]]:
         raise NotImplementedError()
 
+    def _check_vector(self, vector: Union[np.ndarray, float, int]) -> Optional[Union[np.ndarray, float, int]]:
```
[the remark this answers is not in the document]
Yatoom: I agree.

ConfigSpace/hyperparameters.pyx
Yatoom: That's great! I think we can use `np.isnan(a).any()` as it is the fastest and you mentioned below that there are usually no infs/-infs around?

ConfigSpace/hyperparameters.pyx
```diff
@@ -326,18 +323,30 @@ cdef class FloatHyperparameter(NumericalHyperparameter):
...
+        try:
+            if isinstance(vector, np.ndarray):
+                return self._transform_vector(vector)
+            return self._transform_scalar(vector)
+        except ValueError:
+            return None
 
-cdef class IntegerHyperparameter(NumericalHyperparameter):
-    cdef ufhp
+    cpdef double _transform_scalar(self, double scalar):
```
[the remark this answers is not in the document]
Yatoom: If we implement it as a `cdef` can we still call `_transform_scalar()` without calling `_transform()` first?

ConfigSpace/hyperparameters.pyx
Yatoom: Okay sounds good!

ConfigSpace/hyperparameters.pyx
Yatoom: Hmm that's weird... Shouldn't this have given an error?

ConfigSpace/hyperparameters.pyx
Yatoom: Good point.

ConfigSpace/hyperparameters.pyx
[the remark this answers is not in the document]
Yatoom: Actually, do we really need the checks for the vector here? Because all these methods do is to output the `self.value`.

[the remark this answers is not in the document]
Yatoom: Does this make sense?

```python
def _transform(self, vector: Optional[Union[np.ndarray, float, int]]) \
        -> Optional[Union[np.ndarray, float, int]]:
    return self.value

def _transform_vector(self, vector: Optional[np.ndarray]) \
        -> Optional[Union[np.ndarray, float, int]]:
    return self.value

def _transform_scalar(self, vector: Optional[Union[float, int]]) \
        -> Optional[Union[np.ndarray, float, int]]:
    return self.value
```

ConfigSpace/hyperparameters.pyx
```diff
@@ -519,7 +517,7 @@ cdef class UniformFloatHyperparameter(FloatHyperparameter):
         return np.maximum(self.lower, np.minimum(self.upper, vector))
 
     cpdef double _transform_scalar(self, double scalar):
-        if not math.isfinite(scalar):
+        if scalar == np.nan:
```
mfeurer: Unfortunately, that won't work as comparing to `np.NaN` always results in False:
```                                                                                                                                                          
In [3]: a == np.NaN                                                                                                                                                            
Out[3]: False
In [4]: a = 5                                                                                                                                                                  
In [5]: a == np.NaN                                                                                                                                                            
Out[5]: False
```

mfeurer: Instead, you can do:
```
In [8]: a = np.NaN                                                                                                                                                             
In [9]: a != a                                                                                                                                                                 
Out[9]: True
```

ConfigSpace/hyperparameters.pyx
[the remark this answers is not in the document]
mfeurer: Yes, that's fine. It makes the API consistent and keeps the existing behavior.

ConfigSpace/hyperparameters.pyx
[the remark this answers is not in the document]
mfeurer: As long it is called from a `cdef` or `cpdef` function/method. This should be the case, right?

ConfigSpace/hyperparameters.pyx
```diff
@@ -723,7 +721,7 @@ cdef class NormalFloatHyperparameter(FloatHyperparameter):
         return vector
 
     cpdef double _transform_scalar(self, double scalar):
-        if not math.isfinite(scalar):
+        if scalar == np.nan:
```
mfeurer: Same as above.

ConfigSpace/hyperparameters.pyx
mfeurer: Yes, I would have expected that, too.

Yatoom:
Ah, I just realized what "requested changes" actually means. Didn't see the changed code before. Are your changes also included now?

mfeurer:
> Ah, I just realized what "requested changes" actually means. Didn't see the changed code before. Are your changes also included now?

Sorry, I don't get this. Did I change some part of the code?

Yatoom:
Oh nevermind, I thought you did. I got confused because there was a button that said "View Changes".

PR #135, opened 2019-11-18
Introduce weights for categorical hyperparameters

d060030:
This pull requests introduces a new parameter to CategoricalHyperparameter's constructor to assign weights to the given choices. The parameter is optional and, if omitted, the default behavior remains unchanged, i.e. uniform sampling. If set, the weights are normalized and used as probabilities for value sampling. Exceptions are thrown if the number of given choices and given weights differ, if at least one weight is negative or if all weights are zero.


test/test_hyperparameters.py
```diff
@@ -741,6 +741,125 @@ def actual_test():
...
+            choices=[-0.1, 0.0, 0.3],
+            default_value=0.3,
+            weights=[0.1, 0.6, 0.3]
+        )
+        self.assertIn(member=cat_hp_float.sample(rs), container=[-0.1, 0.0, 0.3])
+
+    def test_categorical_with_some_zero_weights(self):
+        # zero weights are okay as long as there is at least one strictly positive weight
+
+        rs = np.random.RandomState(1)
```
mfeurer: I think you need to remove the seed or same more values to make this unit test work as expected. Currently, it will sample the same configuration from the hyperparameter in each iteration. As only one configuration is sampled, changes introducing errors might not be detected.

test/test_hyperparameters.py
```diff
@@ -741,6 +741,125 @@ def actual_test():
...
+            choices=[-0.1, 0.0, 0.3],
+            default_value=0.3,
+            weights=[0.0, 0.6, 0.3]
+        )
+        self.assertIn(member=cat_hp_float.sample(rs), container=[0.0, 0.3])
+
+    def test_categorical_with_all_zero_weights(self):
+        rs = np.random.RandomState(1)
+
+        with self.assertRaises(ValueError) as context:
```
mfeurer: You could simplify the code using `with self.assertRaisesRegex(ValueError, msg): ...`.

test/test_hyperparameters.py
```diff
@@ -741,6 +741,125 @@ def actual_test():
...
+
+        rs = np.random.RandomState(1)
+
+        cat_hp_str = CategoricalHyperparameter(
+            name="param",
+            choices=["A", "B", "C"],
+            default_value="A",
+            weights=[0.1, 0.0, 0.3]
+        )
+        self.assertIn(member=cat_hp_str.sample(rs), container=["A", "C"])
```
mfeurer: You could also explicitly check the probabilities list to be on the safe side.

test/test_hyperparameters.py
d060030: Good point, I check now multiple values and without a fixed seed as no values are checked explicitly.

test/test_hyperparameters.py
d060030: done

test/test_hyperparameters.py
d060030: done

d060030:
Hi Matthias,

thanks for your quick feedback! I fully agree and I tried to incorporate your suggestions. Please let me know if you see room for further improvement.

Best regards,
  Manuel

mfeurer:
This looks great, thanks a lot for your work and quick response!

PR #138, opened 2019-11-28
Fixes #137 

### LINKED ISSUE #137: setup.py fails on system with ASCII default encoding
Hello,
the pip install of ConfigSpace fails on systems with a default encoding not supporting the letter "ü".
This is due to the readfile of the "README" without an explicit encoding. 

As docker container per default use ASCII as their encoding this will not allow installing ConfigSpace inside such a container.
Suggestion: Please include encoding="utf-8" in every open() of a file.

    # Helper functions
    def read_file(fname):
        """Get contents of file from the modules directory"""
        return open(os.path.join(os.path.dirname(__file__), fname), encoding="utf-8").read()

Thank you!

PhMueller:
This should fix the UnicodeDecodeError #137.

UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 1060: ordinal not in range(128)
```bash
>>> open('README.md').read()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.6/encodings/ascii.py", line 26, in decode
    return codecs.ascii_decode(input, self.errors)[0]
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 1060: ordinal not in range(128)`
```

```bash
>>> open('README.md', encoding='utf-8').read()
'# ConfigSpace\n\nA simple python module to  ...
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -8,12 +8,12 @@ from setuptools.command.build_ext import build_ext
 # Helper functions
 def read_file(fname):
     """Get contents of file from the modules directory"""
-    return open(os.path.join(os.path.dirname(__file__), fname)).read()
+    return open(os.path.join(os.path.dirname(__file__), fname), encoding='utf-8').read()
 
 
 def get_version(fname):
     """Get the module version"""
-    with open(fname) as file_handle:
+    with open(fname, encoding='utf-8') as file_handle:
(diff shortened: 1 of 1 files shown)


mfeurer:
Thanks a lot for the fast fix!

PR #142, opened 2020-01-22
Added grid generation utlity for ConfigSpaces

RaghuSpaceRajan:


d060030:
Two minor remarks:
- if the defined, the grid generation could utilize the quantization for UniformFloatHyperparameter and UniformIntegerHyperparameter instead of relying on num_steps_dict
- the function's description could mention that constraints in the config space are ignored when generating the grid

RaghuSpaceRajan:
Thanks for the feedback!

- I did not know about the quantization factor. Would be good to have it as a default step size for the grid.
- I updated the documentation to reflect that conditional HyperParameters are ignored currently. I wanted to add those, but probably in the future at some point.

Neeratyoy:
Hi, is this going to be merged?

RaghuSpaceRajan:
> We should probably have a look at whether that's possible. @RaghuSpaceRajan is there a reason you chose this to be a class and not a function?

On looking at it again, I think it was because there was a function called `get_grid_point` (in my local workspace) to get a **specific point** in the generated grid using a list of indices for the different HPs using their "natural" order in `ConfigSpace`, e.g. passing `[0, 2, 1]` for a grid of `3` HPs would return the grid point containing the 1st HP's 0th grid value, the 2nd HP's 2nd grid value and the 3rd HP's 1st grid value (_0-indexed_). I was storing the generated grid as a member variable to not repeat grid creation every time I wanted to query a particular grid point. But then I added grid generation for conditional spaces and "counting" in the generated grid becomes much more involved and I removed `get_grid_point`.

If you prefer, I can make grid generation be a function again, @mfeurer.

> Also, to have this merged we'd need some unit tests, especially for testing the behavior when a hyperparameter has a given quantization.

I will add these soon.

ConfigSpace/util.pyx
```diff
@@ -436,3 +436,185 @@ def fix_types(configuration: dict,
             else:
                 raise TypeError("Unknown hyperparameter type %s" % type(param))
     return configuration
+
+cimport cython
```
mfeurer, line 440: Could you please put imports at the top of the file?

ConfigSpace/util.pyx
```diff
@@ -436,3 +436,185 @@ def fix_types(configuration: dict,
             else:
                 raise TypeError("Unknown hyperparameter type %s" % type(param))
     return configuration
+
+cimport cython
+@cython.boundscheck(True)  # Activate bounds checking
```
mfeurer, line 441: Great that you figured out how to only change the settings for only this function!

ConfigSpace/util.pyx
```diff
@@ -436,3 +436,185 @@ def fix_types(configuration: dict,
             else:
                 raise TypeError("Unknown hyperparameter type %s" % type(param))
     return configuration
+
+cimport cython
+@cython.boundscheck(True)  # Activate bounds checking
+@cython.wraparound(True)  # Activate negative indexing
+def generate_grid(configuration_space: ConfigurationSpace,
+                 num_steps_dict: Union[None,  Dict[str, int]] = None,
```
mfeurer, line 444: In case you want to simplify this you can use `Optional[Dict[str, int]]`.

ConfigSpace/util.pyx
```diff
@@ -436,3 +436,185 @@ def fix_types(configuration: dict,
...
+    num_steps_dict: dict
+        A dict containing the number of points to divide the grid side formed by Hyperparameters which are either of type UniformFloatHyperparameter or type UniformIntegerHyperparameter. The keys in the dict should be the names of the corresponding Hyperparameters and the values should be the number of points to divide the grid side formed by the corresponding Hyperparameter in to.
+
+    Returns
+    -------
+    list
+        List containing Configurations. It is a cartesian product of tuples of HyperParameter values. Each tuple lists the possible values taken by the corresponding HyperParameter. Within the cartesian product, in each element, the ordering of HyperParameters is the same for the OrderedDict within the ConfigurationSpace.
+    """
+
+    def get_value_set(num_steps_dict, hp_name):
```
mfeurer, line 463: Could you please add type annotations here, too?

ConfigSpace/util.pyx
```diff
@@ -436,3 +436,185 @@ def fix_types(configuration: dict,
...
+                num_steps = num_steps_dict[param.name]
+                grid_points = np.linspace(lower, upper, num_steps)
+            else:
+                grid_points = np.arange(lower, upper + param.q, param.q) # check for log and for rounding issues
+
+            if param.log:
+                grid_points = np.exp(grid_points)
+
+            # Avoiding rounding off issues
+            print(grid_points)
```
mfeurer, line 509: Could you please remove this print call?

ConfigSpace/util.pyx
```diff
@@ -436,3 +436,185 @@ def fix_types(configuration: dict,
...
+    -------
+    list
+        List containing Configurations. It is a cartesian product of tuples of HyperParameter values. Each tuple lists the possible values taken by the corresponding HyperParameter. Within the cartesian product, in each element, the ordering of HyperParameters is the same for the OrderedDict within the ConfigurationSpace.
+    """
+
+    def get_value_set(num_steps_dict, hp_name):
+        '''
+        Gets values along the grid for a particular hyperparameter.
+
+        Uses the num_steps_dict to determine number of grid values for UniformFloatHyperparameter and UniformIntegerHyperparameter. If these values are not present in num_steps_dict, the quantization, q, objects of these classes will be used to divide the grid. NOTE: If q is None, this results in np.arange using None for its step size parameter which results in a quantization of 1.
```
mfeurer, line 467: I'm wondering whether it would make more sense to fail if neither the number of steps nor q is given instead of assuming the user meant to set the value to 1. What do you think?

ConfigSpace/util.pyx
```diff
@@ -436,3 +436,185 @@ def fix_types(configuration: dict,
...
+                grid_points[0] = param.lower
+            if grid_points[-1] > param.upper:
+                grid_points[-1] = param.upper
+
+            return tuple(grid_points)
+
+        else:
+            raise TypeError("Unknown hyperparameter type %s" % type(param))
+
+    def get_cartesian_product(value_sets, hp_names):
```
mfeurer, line 544: Same as above, could you please add type annotations here?

ConfigSpace/util.pyx
```diff
@@ -436,3 +436,185 @@ def fix_types(configuration: dict,
...
+
+        Returns
+        -------
+        list of dicts
+            List of configuration dicts
+
+        '''
+        grid = []
+        import itertools
+        for i, element in enumerate(itertools.product(*value_sets)):
```
mfeurer, line 566: Do you need the `i` for something? If not, could you please drop the `enumerate` here?

ConfigSpace/util.pyx
```diff
@@ -436,3 +436,185 @@ def fix_types(configuration: dict,
...
+    # Create a Cartesian product of above allowed values for the HPs. Hold them in an "unchecked" list because some of the conditionally dependent HPs may become active for some of the elements of the Cartesian product and in these cases creating a Configuration would throw an Error (see below).
+    unchecked_grid_pts = get_cartesian_product(value_sets, hp_names) # Creates a list of Configuration dicts
+    checked_grid_pts = []
+    conditional_grid_lens = []
+
+    while len(unchecked_grid_pts) > 0:
+        try:
+            grid_point = Configuration(configuration_space, unchecked_grid_pts[0])
+            checked_grid_pts.append(grid_point)
+        except ValueError as e:
```
mfeurer, line 592: I'm wondering whether this check is too broad and whether we should add new exceptions for this case?

ConfigSpace/util.pyx
```diff
@@ -436,3 +436,185 @@ def fix_types(configuration: dict,
...
+    value_sets = [] # list of tuples: each tuple within is the grid values to be taken on by a Hyperparameter
+    hp_names = []
+
+    # Get HP names and allowed grid values they can take for the HPs at the top level of ConfigSpace tree
+    for hp_name in configuration_space._children['__HPOlib_configuration_space_root__']:
+        value_sets.append(get_value_set(num_steps_dict, hp_name))
+        hp_names.append(hp_name)
+
+    # Create a Cartesian product of above allowed values for the HPs. Hold them in an "unchecked" list because some of the conditionally dependent HPs may become active for some of the elements of the Cartesian product and in these cases creating a Configuration would throw an Error (see below).
+    unchecked_grid_pts = get_cartesian_product(value_sets, hp_names) # Creates a list of Configuration dicts
```
mfeurer, line 584: Could you please use a double ended queue (`collections.deque`) here to make the code a bit more intuitive?

ConfigSpace/util.pyx
```diff
@@ -436,3 +436,185 @@ def fix_types(configuration: dict,
...
+                            if not cond.evaluate(unchecked_grid_pts[0]):
+                                all_cond_ = False
+                        if all_cond_:
+                            new_active_hp_names.append(new_hp_name)
+
+
+            for hp_name in new_active_hp_names:
+                value_sets.append(get_value_set(num_steps_dict, hp_name))
+                hp_names.append(hp_name)
+            if len(new_active_hp_names) > 0: # this check might not be needed, as there is always going to be a new active HP when in this except block? #TODO
```
mfeurer, line 613: you could add an else that should never be triggered just to be sure?

ConfigSpace/util.pyx
```diff
@@ -436,3 +436,185 @@ def fix_types(configuration: dict,
...
+
+    # Get HP names and allowed grid values they can take for the HPs at the top level of ConfigSpace tree
+    for hp_name in configuration_space._children['__HPOlib_configuration_space_root__']:
+        value_sets.append(get_value_set(num_steps_dict, hp_name))
+        hp_names.append(hp_name)
+
+    # Create a Cartesian product of above allowed values for the HPs. Hold them in an "unchecked" list because some of the conditionally dependent HPs may become active for some of the elements of the Cartesian product and in these cases creating a Configuration would throw an Error (see below).
+    unchecked_grid_pts = get_cartesian_product(value_sets, hp_names) # Creates a list of Configuration dicts
+    checked_grid_pts = []
+    conditional_grid_lens = []
```
mfeurer, line 586: Could you briefly explain what this variable is used for?

test/test_util.py
```diff
@@ -342,3 +342,74 @@ def test_fix_types(self):
...
+        ord1 = OrdinalHyperparameter(name='ord1', sequence=['1', '2', '3'])
+
+        cs.add_hyperparameters([float1, int1, cat1, ord1, const1])
+
+        num_steps_dict = {'float1': 11, 'int1': 6}
+        generated_grid = generate_grid(cs, num_steps_dict)
+
+        # Check randomly pre-selected values in the generated_grid
+        # 2 * 1 * 11 * 6 * 3 total diff. possible configurations
+        self.assertEqual(len(generated_grid), 396)
```
mfeurer, line 363: Hey, how about you check a few configurations, such as the first and the last one?

test/test_util.py
```diff
@@ -342,3 +342,74 @@ def test_fix_types(self):
...
+        self.assertEqual(generated_grid[0].get_dictionary()['cat1'], 'T')
+        self.assertEqual(generated_grid[198].get_dictionary()['cat1'], 'F')
+        self.assertEqual(generated_grid[45].get_dictionary()['const1'], 4)
+        # The 2 most frequently changing HPs (int1 and ord1) have 3 * 6 = 18 different values for
+        # each value of float1, so the 4th value of float1 of -0.4 is reached after
+        # 3 * 18 = 54 values in the generated_grid (and remains the same for the next 18 values):
```
mfeurer, line 369: If it remains the same for the next 18 values, could you maybe check that?

test/test_util.py
```diff
@@ -342,3 +342,74 @@ def test_fix_types(self):
...
+        '''Test grid generation'''
+        cs = ConfigurationSpace(seed=1234)
+
+        cat1 = CategoricalHyperparameter(name='cat1', choices=['T', 'F'])
+        const1 = Constant(name='const1', value=4)
+        float1 = UniformFloatHyperparameter(name='float1', lower=-1, upper=1, log=False)
+        int1 = UniformIntegerHyperparameter(name='int1', lower=10, upper=100, log=True)
+        ord1 = OrdinalHyperparameter(name='ord1', sequence=['1', '2', '3'])
+
+        cs.add_hyperparameters([float1, int1, cat1, ord1, const1])
```
mfeurer, line 356: Could you please check for a few extreme cases, too? Such as only numerical, only categorical, only constant, only a single hyperparameter, no hyperparameters yet?

test/test_util.py
```diff
@@ -342,3 +342,74 @@ def test_fix_types(self):
...
+                                GreaterThanCondition(cat1_cond, float1, -0.5),
+                                LessThanCondition(cat1_cond, float1, 0.5)
+                                )
+        cs2.add_conditions([cond_2])
+        float2_cond = UniformFloatHyperparameter(name='float2_cond',
+                                                 lower=10., upper=100., log=True)
+        cs2.add_hyperparameters([float2_cond])
+        cond_3 = GreaterThanCondition(float2_cond, int2_cond, 50)
+        cs2.add_conditions([cond_3])
+        # print(cs2)
```
mfeurer, line 403: Could you please drop the commented print statements?

test/test_util.py
```diff
@@ -342,3 +342,74 @@ def test_fix_types(self):
...
+        cs2.add_conditions([cond_1])
+        cat1_cond = CategoricalHyperparameter(name='cat1_cond', choices=['apple', 'orange'])
+        cs2.add_hyperparameters([cat1_cond])
+        cond_2 = AndConjunction(GreaterThanCondition(cat1_cond, int1, 300),
+                                LessThanCondition(cat1_cond, int1, 700),
+                                GreaterThanCondition(cat1_cond, float1, -0.5),
+                                LessThanCondition(cat1_cond, float1, 0.5)
+                                )
+        cs2.add_conditions([cond_2])
+        float2_cond = UniformFloatHyperparameter(name='float2_cond',
```
mfeurer, line 398: Could you please add a comment that this is a 2nd-level dependency so that it's immediately clear that this is tested?

test/test_util.py
```diff
@@ -342,3 +342,74 @@ def test_fix_types(self):
...
+        # print(cs2)
+        num_steps_dict1 = {'float1': 4, 'int2_cond': 3, 'float2_cond': 3}
+        generated_grid = generate_grid(cs2, num_steps_dict1)
+        # print(generated_grid)
+        self.assertEqual(len(generated_grid), 18)
+
+        # RR: I manually generated the grid and verified the values were correct.
+        # Here, we test that a few randomly chosen values in the generated grid
+        # correspond to the ones I checked.
+        self.assertEqual(generated_grid[3].get_dictionary()['int1'], 1000)
```
mfeurer, line 412: As above, could you please check maybe the first and last configuration completely?

ConfigSpace/util.pyx
RaghuSpaceRajan: Done

ConfigSpace/util.pyx
RaghuSpaceRajan: Yes, the link here: [https://cython.readthedocs.io/en/latest/src/userguide/numpy_tutorial.html#tuning-indexing-further](https://cython.readthedocs.io/en/latest/src/userguide/numpy_tutorial.html#tuning-indexing-further) seems to suggest array lookups _are_ slowed down by deactivating those.

ConfigSpace/util.pyx
RaghuSpaceRajan: Done

ConfigSpace/util.pyx
RaghuSpaceRajan: Done

ConfigSpace/util.pyx
RaghuSpaceRajan: Yes, I had thought about throwing a warning, especially if the number of possible grid points is "too many". But probably, an Error is better. Added that now.

ConfigSpace/util.pyx
RaghuSpaceRajan: Done

ConfigSpace/util.pyx
RaghuSpaceRajan: Done

ConfigSpace/util.pyx
RaghuSpaceRajan: I could check for the error message in `ValueError` to match `Active hyperparameter '%s' not specified!` and print out an info/warning message with the `ValueError` that was caught so that it is not missed. And when the text doesn't match we could rethrow the error.

ConfigSpace/util.pyx
RaghuSpaceRajan: Done

ConfigSpace/util.pyx
RaghuSpaceRajan: And throw an Error inside that else?

ConfigSpace/util.pyx
RaghuSpaceRajan: It was more for debug purposes, to check the lengths of "sub-grids" generated in different parts of a conditional `ConfigurationSpace`. I have removed it now.

test/test_util.py
RaghuSpaceRajan: Done

test/test_util.py
RaghuSpaceRajan: Done

test/test_util.py
RaghuSpaceRajan: Done

ConfigSpace/util.pyx
RaghuSpaceRajan: It still seems to require specifying the default of `None` though.

test/test_util.py
RaghuSpaceRajan: For the case of no hyperparameters, in `get_cartesian_product`, `itertools.product()` returns a single empty tuple element (I expected 0 elements) here: [https://github.com/RaghuSpaceRajan/ConfigSpace/blob/19144c34b4b0cce2e9df1bb66943bbd17392c818/ConfigSpace/util.pyx#L566]() which leads to a single empty Configuration being generated in the grid. Should we check for this and make sure no configuration is returned?

test/test_util.py
RaghuSpaceRajan: It was mentioned above in the code. I mentioned it again here.

test/test_util.py
RaghuSpaceRajan: Done

ConfigSpace/util.pyx
mfeurer: Yes, exactly.

ConfigSpace/util.pyx
mfeurer: I think raising if the text doesn't match is sufficient. No need to print a warning/info if it matches.

test/test_util.py
mfeurer: Great, could you do me another favor and compare the whole dictionary at once? That would make the code more readable.

test/test_util.py
mfeurer: > Should we check for this and make sure no configuration is returned?

Yes, I think so.

ConfigSpace/util.pyx
RaghuSpaceRajan: Done.

ConfigSpace/util.pyx
RaghuSpaceRajan: Done.

test/test_util.py
RaghuSpaceRajan: Done.

mfeurer (approved):

PR #145, opened 2020-03-29
Fixed "probabilities" can't copy bug in "add_configuration_space"

tqichun:
in configuration_space.pyx#add_configuration_space, if add child configuration_space to parent, the "probabilities" attribute will disappear, in another way, "probabilities" can't copy bug in "add_configuration_space" function. To fix it , I use "copy.deepcopy" instead of "copy.copy" in ConfigSpace/configuration_space.pyx:637 .


mfeurer:
Thanks a lot for you PR. However, I think there's a simpler method to achieve this. Each hyperparameter class has a method `__copy__`, and if you add the probabilities in there, it'll copy them, too.

mfeurer:
Thanks for your effort, but we just merged the easier fix in #149.

PR #146, opened 2020-04-30
Add Callable support for CategoricalHyperparameter

### LINKED ISSUE #133: Allow callbacks or objects to be passed to `CategoricalHyperparameter`
Currently there is no way to pass an object (not `int`, `float` or `str`) when declaring a set of `CategoricalHyperparameter`.

Would be very nice to be able to do something like: ```CategoricalHyperparameter('final_activation', [nn.Softmax(dim=1), nn.Sigmoid()])```

Or a callable like ```CategoricalHyperparameter('final_activation', [lambda : nn.Softmax(dim=4)])```

Currently I have to pass a string saying 'softmax', check for that value in the model and create corresponding layer - this would allow to be less verbose while gaining in simplicity.

ThibzR:
fix #133 

One can now directly pass a Callable when creating a `CategoricalHyperparameter`. This is streamlining prototyping. Couple of examples:

- `CategoricalHyperparameter('final_activation', [nn.Softmax(dim=1), nn.Sigmoid()])`
- `CategoricalHyperparameter('loss_function', [nn.L1Loss, nn.MSELoss])`

Previously, one needed to provide strings, check value later and create the object accordingly.

I've checked my changes on a heavy hyperparameter optimization project of mine and works like a charm. Let me know if any comment!

### DIFF
--- a/ConfigSpace/configuration_space.pyx
+++ b/ConfigSpace/configuration_space.pyx
@@ -52,7 +52,7 @@ from ConfigSpace.forbidden import (
     AbstractForbiddenClause,
     AbstractForbiddenConjunction,
 )
-from typing import Union, List, Any, Dict, Iterable, Set, Tuple, Optional
+from typing import Union, List, Any, Dict, Iterable, Set, Tuple, Optional, Callable
 from ConfigSpace.exceptions import ForbiddenValueError
 import ConfigSpace.c_util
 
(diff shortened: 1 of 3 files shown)


shukon:
I'm not maintaining this project, but my two cents: how will this behave on save-and-reload in e.g. json?

mfeurer:
> I'm not maintaining this project, but my two cents: how will this behave on save-and-reload in e.g. json?

It would basically make it impossible to save/load configuration spaces expect for using pickle.

I hope that I can have a look at this PR soonish.

ThibzR:
> > I'm not maintaining this project, but my two cents: how will this behave on save-and-reload in e.g. json?
> 
> It would basically make it impossible to save/load configuration spaces expect for using pickle.

Indeed, pickle handles that nicely. I'd say the feature comes with a tradeoff, but no breaking change as to the existing package.

eddiebergman:
Sorry, this was a nice suggestoin and thanks for the PR! It's been superseded by #359 and allows arbitrary objects, including functions, in the categorical or ordinal :) Will close and leave the original issue open until that PR is merged!

PR #167, opened 2020-10-17
Fix forbidden Configurations Space

franchuterivera:
I found a typo in the config space comparison, which would be great to fix.

But more importantly, in the parallel context objects might be re-created. There is an if below that basically sets up a value maybe if this is the first time a function is called. This has to be done separately, as the "other" config might be re-created while te "self" is not.


mfeurer (approved):
Looks reasonable to me.

PR #172, opened 2021-01-12
Fix problems in github actions

franchuterivera:
This PR is Intended to fix the problems in github actions:

- Flake fixes
- Try to reproduce docs error

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -21,7 +21,7 @@ jobs:
       if: (contains(github.ref, 'master')) && github.event_name == 'push'
       run: |
         cd ..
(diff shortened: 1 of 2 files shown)


mfeurer:
> Try to reproduce docs error

To do so, we could actually run the checkout and copy for all documentation builds and only the upload for documentation builds on the master. This would allow for easier debugging of the doc building process. What do you think?

franchuterivera:
It was a problem with the https! I forced a push in my repo here https://github.com/franchuterivera/ConfigSpace/runs/1688614899?check_suite_focus=true 

So when I did the clone, I was cloning via ssh which apparently is not supported.

Issue #173, opened 2021-02-01
Problem when importing ConfigSpace (0.4.17)


drake666:
Today, when I've tried importing the ConfigSpace package on Ubuntun 18.04, with python 3.8.0, the action raised a ValueError:

```
Python 3.8.0 (default, Oct 28 2019, 16:14:01) 
[GCC 8.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import ConfigSpace
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/maxime/pipe-inspection/machine_learning/training/.venv/lib/python3.8/site-packages/ConfigSpace/__init__.py", line 37, in <module>
    from ConfigSpace.configuration_space import Configuration, \
  File "ConfigSpace/configuration_space.pyx", line 39, in init ConfigSpace.configuration_space
  File "ConfigSpace/hyperparameters.pyx", line 1, in init ConfigSpace.hyperparameters
ValueError: numpy.ndarray size changed, may indicate binary incompatibility. Expected 88 from C header, got 80 from PyObject

```

I'm currently using numpy 1.19.5 (can't use 1.20.0 since I'm also using tensorflow 2.4.0). Am I alone with this predicament?

jbedorf:
I ran into the same issue. The issue is caused by the release of numpy 1.20.0 a few days ago. The installation of ConfigSpace does not restrict which numpy version that is used during build process, and as such the latest version is downloaded during the compilation of the ConfigSpace package. Which means that ConfigSpace is compiled against numpy 1.20.0 which has a different PyObject size than the pre 1.20 versions, and the one installed in your environment for usage.

I've found two ways to work around it for now:

1. Disable pip's build isolation 

Run pip install with the 'no-build-isolation' flag. This way pip will use your already installed numpy version: "pip install --no-build-isolation ConfigSpace==0.4.16"

`pip install --no-build-isolation ConfigSpace==0.4.16`

2. Edit the package source

Clone the repository, and modify, the [pyproject.toml](https://github.com/automl/ConfigSpace/blob/master/pyproject.toml) file to read:  

```
[build-system]
requires = ["setuptools", "wheel", "numpy < 1.20.0", "Cython"]
```
Then create/install the package. (python setup.py sdist ; pip install dist/...)

benjamc:
Same problem here. 

Found a solution:
First, uninstall ConfigSpace
`pip uninstall ConfigSpace`
Then do following steps:
[...]

PR #175, opened 2021-02-11
Update pyproject.toml

stradigi-mario-bruno:
Fix for [#173](https://github.com/automl/ConfigSpace/issues/173)
You can read more about it here https://github.com/scipy/oldest-supported-numpy

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,2 +1,2 @@
 [build-system]
-requires = ["setuptools", "wheel", "numpy", "Cython"]
+requires = ["setuptools", "wheel", "oldest-supported-numpy", "Cython"]


leezu:
Could there still be ABI issues in edge cases where the runtime environment contains an optimized numpy version, or for example a numpy debug build? Instead of trying to find workarounds to ensure that the build and runtime version are ABI compatible, I think the cleanest solution here is to just use the same version of numpy during build and runtime by not opting-in to the build isolation feature via https://github.com/facebookresearch/fairscale/commit/e92e85ceb5bb1352e2b4a0f645275f181a159018

stradigi-mario-bruno:
> Could there still be ABI issues in edge cases where the runtime environment contains an optimized numpy version, or for example a numpy debug build? Instead of trying to find workarounds to ensure that the build and runtime version are ABI compatible, I think the cleanest solution here is to just use the same version of numpy during build and runtime by not opting-in to the build isolation feature via [facebookresearch/fairscale@e92e85c](https://github.com/facebookresearch/fairscale/commit/e92e85ceb5bb1352e2b4a0f645275f181a159018)

This is also a good solution, thanks! I guess we'll leave it up to the maintainers to decide the best option for them.

gradientsky:
+1 on applying ANY fix: this should resolve downstream issues with isolated builds:
```
ValueError: numpy.ndarray size changed, may indicate binary incompatibility. Expected 88 from C header, got 80 from PyObject
```

PR #182, opened 2021-04-21
[ADD] Move from richcmp to __eq__()

franchuterivera:
Enables the use of __eq__() as it is now supported in cython.

### DIFF
--- a/ConfigSpace/conditions.pyx
+++ b/ConfigSpace/conditions.pyx
@@ -129,7 +129,7 @@ cdef class AbstractCondition(ConditionComponent):
 
         """
         if not isinstance(other, self.__class__):
-            return NotImplemented
+            return False
 
         if self.child != other.child:
             return False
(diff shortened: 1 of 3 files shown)


ConfigSpace/conditions.pyx
```diff
@@ -115,41 +115,27 @@ cdef class AbstractCondition(ConditionComponent):
...
+            unless it is NotImplemented.
 
-            elif op == 3:
-                if (self.child != other.child or
-                        self.parent != other.parent or self.value != other.value):
-                    return True
-                else:
-                    return False
+        """
+        if not isinstance(other, self.__class__):
```
mfeurer, line 131: Sorry, I don't understand this. Shouldn't this simply be `return False`?

ConfigSpace/conditions.pyx
```diff
@@ -545,41 +531,28 @@ cdef class InCondition(AbstractCondition):
         self.value = values
         self.vector_values = [self.parent._inverse_transform(value) for value in self.values]
 
-    def __richcmp__(self, other: Any, int op):
+    def __eq__(self, other: Any) -> bool:
```
mfeurer, line 534: I'm currently wondering why we need this given that this is a subclass of the `AbstractCondition`

ConfigSpace/conditions.pyx
```diff
@@ -618,47 +591,31 @@ cdef class AbstractConjunction(ConditionComponent):
...
-                    raise NotImplemented
-            return True
+        if not isinstance(other, self.__class__):
+            return NotImplemented
+
+        if len(self.components) != len(other.components):
+            return False
+
+        for component, other_component in zip(self.components, other.components):
+            if not (component == other_component):
```
mfeurer, line 614: How about `!=` to simplify reading this?

ConfigSpace/forbidden.pyx
```diff
@@ -57,39 +57,29 @@ cdef class AbstractForbiddenComponent(object):
...
+        Additionally, it defines the __ne__() as stated in the
+        documentation from python:
+            By default, object implements __eq__() by using is, returning NotImplemented 
+            in the case of a false comparison: True if x is y else NotImplemented. 
+            For __ne__(), by default it delegates to __eq__() and inverts the result 
+            unless it is NotImplemented.
+
         """
+        if not isinstance(other, self.__class__):
+            return NotImplemented
```
mfeurer, line 74: I guess my comment from above holds at all of these occurences.

mfeurer:
Hey, do you have any idea why the distcheck fails?

franchuterivera:
Sadly I do not known, as in my laptop (and previous check) that was not a problem. Can you please rerun the check (maybe github actions issue)? I do not think I have permission to do so.

mfeurer:
Alright, today the restarting worked.

PR #183, opened 2021-05-20
[FIX] Cleanup ConfigSpace warnings

franchuterivera:
This PR aims to remove every warning from configspace (so we have a full green pytest run).

This is done by:

- Removing pyparsing warnings
`/home/chico/.local/lib/python3.8/site-packages/pyparsing.py:3190: FutureWarning: Possible set intersection at position 3
`
- Removing the deprecation warnings from numpy
```
<google-sheets-html-origin><style type="text/css"><!--br {mso-data-placement:same-cell;}--></style>
Config space deprecation warnings regarding np.float/np.bool
--
cated in NumPy 1.20; for more details and guidance: https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations
cationWarning: `np.float` is a deprecated alias for the builtin `float`. To silence this warning, use `float` by itself. Doing this will not modify any behavior and is safe. If you specifically wanted    the numpy scalar type, use `np.float64` here.
cationWarning: `np.int` is a deprecated alias for the builtin `int`. To silence this warning, use `int` by itself. Doing this will not modify any behavior and is safe. When replacing `np.int`, you may    wish to use e.g. `np.int64` or `np.int32` to specify the precision. If you wish to review your current use, check the release note link for additional information.
rootdir: /home/chico/work/ConfigSpace_fork_removedeprecation

</google-sheets-html-origin>
```

### DIFF
--- a/ConfigSpace/conditions.pyx
+++ b/ConfigSpace/conditions.pyx
@@ -46,7 +46,7 @@ from ConfigSpace.hyperparameters cimport Hyperparameter
 # We now need to fix a datatype for our arrays. I've used the variable
 # DTYPE for this, which is assigned to the usual NumPy runtime
 # type info object.
-# DTYPE = np.float
+# DTYPE = float
 # "ctypedef" assigns a corresponding compile-time type to DTYPE_t. For
 # every type in the numpy module there's a corresponding compile-time
 # type with a _t-suffix.
--- a/ConfigSpace/conditions.pyx
+++ b/ConfigSpace/conditions.pyx
@@ -43,14 +43,6 @@ import numpy as np
 from ConfigSpace.hyperparameters import NumericalHyperparameter, OrdinalHyperparameter
 from ConfigSpace.hyperparameters cimport Hyperparameter
 
-# We now need to fix a datatype for our arrays. I've used the variable
-# DTYPE for this, which is assigned to the usual NumPy runtime
-# type info object.
-# DTYPE = float
-# "ctypedef" assigns a corresponding compile-time type to DTYPE_t. For
-# every type in the numpy module there's a corresponding compile-time
-# type with a _t-suffix.
-# ctypedef np.float_t DTYPE_t
 cimport numpy as np
 
 
--- a/ConfigSpace/forbidden.pyx
+++ b/ConfigSpace/forbidden.pyx
@@ -37,14 +37,6 @@ from typing import List, Dict, Any, Union
 
 from ConfigSpace.forbidden cimport AbstractForbiddenComponent
 
-# We now need to fix a datatype for our arrays. I've used the variable
-# DTYPE for this, which is assigned to the usual NumPy runtime
-# type info object.
-# DTYPE = float
-# "ctypedef" assigns a corresponding compile-time type to DTYPE_t. For
-# every type in the numpy module there's a corresponding compile-time
-# type with a _t-suffix.
-# ctypedef float_t DTYPE_t
 from libc.stdlib cimport malloc, free
 cimport numpy as np
 
--- a/ConfigSpace/forbidden.pyx
+++ b/ConfigSpace/forbidden.pyx
@@ -40,11 +40,11 @@ from ConfigSpace.forbidden cimport AbstractForbiddenComponent
 # We now need to fix a datatype for our arrays. I've used the variable
 # DTYPE for this, which is assigned to the usual NumPy runtime
 # type info object.
-# DTYPE = np.float
+# DTYPE = float
 # "ctypedef" assigns a corresponding compile-time type to DTYPE_t. For
 # every type in the numpy module there's a corresponding compile-time
 # type with a _t-suffix.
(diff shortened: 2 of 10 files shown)


ConfigSpace/conditions.pyx
```diff
@@ -46,7 +46,7 @@ from ConfigSpace.hyperparameters cimport Hyperparameter
 # We now need to fix a datatype for our arrays. I've used the variable
 # DTYPE for this, which is assigned to the usual NumPy runtime
 # type info object.
-# DTYPE = np.float
+# DTYPE = float
```
mfeurer, line 49: I guess we can drop that commented piece of code?

ConfigSpace/forbidden.pyx
```diff
@@ -40,11 +40,11 @@ from ConfigSpace.forbidden cimport AbstractForbiddenComponent
 # We now need to fix a datatype for our arrays. I've used the variable
 # DTYPE for this, which is assigned to the usual NumPy runtime
 # type info object.
-# DTYPE = np.float
+# DTYPE = float
```
mfeurer, line 43: Can we drop that stuff here, too?

ConfigSpace/hyperparameters.pyx
```diff
@@ -40,7 +40,7 @@ cimport numpy as np
 # We now need to fix a datatype for our arrays. I've used the variable
 # DTYPE for this, which is assigned to the usual NumPy runtime
 # type info object.
-# DTYPE = np.float
+# DTYPE = float
```
mfeurer, line 43: And here?

ConfigSpace/hyperparameters.pyx
```diff
@@ -877,7 +877,7 @@ cdef class UniformIntegerHyperparameter(IntegerHyperparameter):
         if self.log:
             repr_str.write(", on log-scale")
         if self.q is not None:
-            repr_str.write(", Q: %s" % repr(np.int(self.q)))
+            repr_str.write(", Q: %s" % repr(np.int64(self.q)))
```
mfeurer, line 880: Do we have a test for this change?

ConfigSpace/read_and_write/pcs_new.py
```diff
@@ -64,7 +64,9 @@
 pp_number = pp_e_notation | pp_float | pp_int
 pp_numberorname = pp_number | pp_param_name
 pp_log = pyparsing.Word("log")
-pp_connective = pyparsing.Word("||" + "&&")
+# A word matches each character as a set. So && is processed as &
+# https://pythonhosted.org/pyparsing/pyparsing.Word-class.html
+pp_connective = pyparsing.Word("|" + "&")
```
mfeurer, line 69: Shouldn't this be `pyparsing.Literal` then?

mfeurer: And the `log` above, too?

ConfigSpace/hyperparameters.pyx
franchuterivera: Thanks, I just added a test for this.

ConfigSpace/read_and_write/pcs_new.py
franchuterivera: Thanks for the catch of `log`. I have fixed that.

Leaving `pp_connective = pyparsing.Word("|" + "&")` as it is was intentional, as [this line](https://github.com/automl/ConfigSpace/blob/b2fe51e7f03cc080640093d30bbaaa76a11a3b59/ConfigSpace/read_and_write/pcs_new.py#L80) assumes you will be checking for `&` or `|` not `&|`.

I do not know much about pyparsing, but probably we can do some `Combine` or `And` to create this behaviour. In other words, if one just makes this literal [this](https://github.com/franchuterivera/ConfigSpace/blob/63b781a031d6768566001932b0749f82df6caf6d/test/read_and_write/test_pcs_converter.py#L492) unit test fails.

Let me know how you want to proceed (that is re-writing the pyparsing statement or keeping the `pyparsing.Word("|" + "&")`).

mfeurer: That makes sense. Can you check whether there's a union of literals we could use here to rewrite this statement to check for either `"??"` or `"||"`?

ConfigSpace/hyperparameters.pyx
mfeurer: Ha, I'm wondering why we cast here in the first place. We don't cast for upper and lower, so this seems superfluous, what do you think?

ConfigSpace/read_and_write/pcs_new.py
mfeurer: The change looks good. There's no need to keep this simple, this should be used only very rarely anyways, so it can be slow...

mfeurer (approved):
Looks good to me. Is this ready to be merged?

franchuterivera:
Yes, it is, thanks for taking a look.

PR #184, opened 2021-05-20
Enable wheels for 32/64 architecture

franchuterivera:
- Enable wheels on a similar fashion as pyrfr
- Fixes an overflow for 32 bits systems --> np.int is too big to fix c-type long. So I am moving to long long type for the below operations.


franchuterivera:
This PR has an extra line that we need to remove ([this](https://github.com/franchuterivera/ConfigSpace/blob/5f89a5650b8d864a8307cf1deaa49159f796a018/.github/workflows/wheels.yml#L8)) so that I can check your comments :D

Before merging we should remove it

mfeurer (commented):
Looks good to me. Could you maybe add a comment on top of the wheel workflow file describing how to trigger the wheel building action?

franchuterivera:
Please let me know if this message Is informative enough

.github/workflows/wheels.yml
```diff
@@ -1,11 +1,22 @@
 # Workflow to build and test wheels
+# =================================
+# This github action gets triggered whenever there
+# is a push to the master branch. It generates both
+# wheels and distributions files, making sure their
+# contents are correct via unit-testing.
+#
+# However, artifacts are only uploaded to a release
+# here: https://github.com/automl/ConfigSpace/releases
+# When a release tag is created. For example, if
```
mfeurer, line 10: Hey, could you please be more specific here? In particular, does it have to be a release (I think that's github nomenclature) or a tag (that's git nomenclature)? I think it would also be good if you could briefly add the steps that are required to be done from the command line.

mfeurer (approved):

PR #188, opened 2021-07-26
support for truncated normal distribution

pbalapra:
Adding support for truncated normal distribution for NormalIntegerHyperparameter and NormalFloatHyperparameter with lower and upper bound.

### DIFF
--- a/ConfigSpace/hyperparameters.pyx
+++ b/ConfigSpace/hyperparameters.pyx
@@ -433,7 +433,7 @@ cdef class UniformFloatHyperparameter(FloatHyperparameter):
         ----------
         name : str
             Name of the hyperparameter, with which it can be accessed
-        lower : int, floor
+        lower : int, float
             Lower bound of a range of values from which the hyperparameter will be sampled
         upper : int, float
             Upper bound
@@ -635,6 +635,10 @@ cdef class NormalFloatHyperparameter(FloatHyperparameter):
         ----------
         name : str
             Name of the hyperparameter, with which it can be accessed
+        lower : int, float
+            Lower bound of a range of values from which the hyperparameter will be sampled
+        upper : int, float
+            Upper bound
         mu : int, float
             Mean of the distribution
         sigma : int, float
--- a/ConfigSpace/hyperparameters.pyx
+++ b/ConfigSpace/hyperparameters.pyx
@@ -35,6 +35,7 @@ from collections import OrderedDict, Counter
 from typing import List, Any, Dict, Union, Set, Tuple, Optional
 
 import numpy as np
+from scipy.stats import truncnorm
 cimport numpy as np
 
 
@@ -610,6 +611,8 @@ cdef class NormalFloatHyperparameter(FloatHyperparameter):
     def __init__(self, name: str, mu: Union[int, float], sigma: Union[int, float],
                  default_value: Union[None, float] = None,
                  q: Union[int, float, None] = None, log: bool = False,
+                 lower: Union[int, None] = None,
+                 upper: Union[int, None] = None,
                  meta: Optional[Dict] = None) -> None:
         r"""
         A float hyperparameter.
@@ -655,6 +658,47 @@ cdef class NormalFloatHyperparameter(FloatHyperparameter):
         self.default_value = self.check_default(default_value)
         self.normalized_default_value = self._inverse_transform(self.default_value)
 
+        if lower is not None and upper is not None:
+            self.lower = float(lower)
+            self.upper = float(upper)
+
+            if self.lower >= self.upper:
+                raise ValueError("Upper bound %f must be larger than lower bound "
+                                "%f for hyperparameter %s" %
+                                (self.upper, self.lower, name))
+            elif log and self.lower <= 0:
+                raise ValueError("Negative lower bound (%f) for log-scale "
+                                "hyperparameter %s is forbidden." %
+                                (self.lower, name))
+
+            self.default_value = self.check_default(default_value)
+
+            if self.log:
+                if self.q is not None:
+                    lower = self.lower - (np.float64(self.q) / 2. + 0.0001)
+                    upper = self.upper + (np.float64(self.q) / 2. - 0.0001)
+                else:
+                    lower = self.lower
+                    upper = self.upper
+                self._lower = np.log(lower)
+                self._upper = np.log(upper)
+            else:
+                if self.q is not None:
+                    self._lower = self.lower - (self.q / 2. + 0.0001)
+                    self._upper = self.upper + (self.q / 2. - 0.0001)
+                else:
+                    self._lower = self.lower
+                    self._upper = self.upper
+            if self.q is not None:
+                # There can be weird rounding errors, so we compare the result against self.q, see
+                # In [13]: 2.4 % 0.2
+                # Out[13]: 0.1999999999999998
+                if np.round((self.upper - self.lower) % self.q, 10) not in (0, self.q):
+                    raise ValueError(
+                        'Upper bound (%f) - lower bound (%f) must be a multiple of q (%f)'
+                        % (self.upper, self.lower, self.q)
+                    )
+
     def __repr__(self) -> str:
         repr_str = io.StringIO()
         repr_str.write("%s, Type: NormalFloat, Mu: %s Sigma: %s, Default: %s" %
@@ -689,7 +733,9 @@ cdef class NormalFloatHyperparameter(FloatHyperparameter):
             self.mu == other.mu and
             self.sigma == other.sigma and
             self.log == other.log and
-            self.q == other.q
+            self.q == other.q and 
+            self.lower == other.lower and 
+            self.upper == other.upper
         )
 
     def __copy__(self):
@@ -700,6 +746,8 @@ cdef class NormalFloatHyperparameter(FloatHyperparameter):
             sigma=self.sigma,
             log=self.log,
             q=self.q,
+            lower=self.lower, 
+            upper=self.upper,
             meta=self.meta
         )
 
@@ -707,9 +755,16 @@ cdef class NormalFloatHyperparameter(FloatHyperparameter):
         return hash((self.name, self.mu, self.sigma, self.log, self.q))
 
     def to_uniform(self, z: int = 3) -> 'UniformFloatHyperparameter':
+        if self.lower is not None and self.upper is not None: 
+            lb = self.mu - (z * self.sigma)
+            ub = self.mu + (z * self.sigma)
+        else:
+            lb = self.lower
+            ub = self.upper
+
         return UniformFloatHyperparameter(self.name,
-                                          self.mu - (z * self.sigma),
-                                          self.mu + (z * self.sigma),
+                                          lb,
+                                          ub,
                                           default_value=int(
                                               np.round(self.default_value, 0)),
                                           q=self.q, log=self.log)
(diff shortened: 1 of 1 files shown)


Deathn0t:
Hello @mfeurer , what do you think about this PR? We found it extremely useful to have hyper-parameters with normal distribution and boundary constraints.

ConfigSpace/hyperparameters.pyx
```diff
@@ -610,6 +611,8 @@ cdef class NormalFloatHyperparameter(FloatHyperparameter):
     def __init__(self, name: str, mu: Union[int, float], sigma: Union[int, float],
                  default_value: Union[None, float] = None,
                  q: Union[int, float, None] = None, log: bool = False,
+                 lower: Union[int, None] = None,
+                 upper: Union[int, None] = None,
```
eddiebergman, line 615: I think it'd be good to go with the explicit `Optional[int]` type here. By definition `Union[int, None] == Optional[int]` but Optional is a bit more syntactically clear

ConfigSpace/hyperparameters.pyx
```diff
@@ -700,16 +751,25 @@ cdef class NormalFloatHyperparameter(FloatHyperparameter):
...
         return hash((self.name, self.mu, self.sigma, self.log, self.q))
 
     def to_uniform(self, z: int = 3) -> 'UniformFloatHyperparameter':
+        if self.lower is not None and self.upper is not None:
+            lb = self.mu - (z * self.sigma)
+            ub = self.mu + (z * self.sigma)
+        else:
+            lb = self.lower
+            ub = self.upper
+
```
eddiebergman, line 769: One or both of `self.lower` and `self.upper` are None here. I imagine the if statement should read:
```

File: ConfigSpace/hyperparameters.pyx [hunk 3]
```diff
@@ -1059,6 +1132,10 @@ cdef class NormalIntegerHyperparameter(IntegerHyperparameter):
         ----------
         name : str
             Name of the hyperparameter with which it can be accessed
+        lower : int
+            Lower bound of a range of values from which the hyperparameter will be sampled
+        upper : int
+            upper bound
         mu : int
```
eddiebergman, line 1139: Similarly here with the type signatures and parameter orders.

eddiebergman (changes requested):
Hi @pbalapra ,

These changes look good and don't intrude on any current functionality. However I think tests are much required for this to ensure it works correctly in all cases. Some interesting test cases which should be tested against:
* How do both of these act when `upper == lower`, are the both inclusive bounds?
* I assume from your usage, the quantization and transforming to uniform works but we would needs tests to confirm that this functionality will always work as intended.

I think the main doc string for both classes could do with a simple line describing that bounds can be set. You can see how the docstrings are rendered [here](https://automl.github.io/ConfigSpace/master/API-Doc.html#ConfigSpace.hyperparameters.NormalFloatHyperparameter).

Deathn0t:
Hello @eddiebergman,

I think I took into consideration all your comments. I added the test in the `test_hyperparameters.py` file. All tests are passing for me with pytest. Let me know if everything is ok.

eddiebergman:
Hi @Deathn0t,

All looks good to me, thanks for adding the tests! The only last small change required is that one of the [tests ](https://github.com/automl/ConfigSpace/pull/188/checks?check_run_id=3602132920)`pre-commit` failed. This is just a code standard checker for which we use `flake8`.

You can either manually fix this and push, I'll rerun the tests and you can see if anything else needs to be fixed. Alternatively, to run this check locally.
```
# Using pre-commit
pip install pre-commit
pre-commit run --all-files
```

You can also just use a `flake8` linter if it's readily available in whatever editor you use, the results should be the same.

 @mfeurer if you want to have a final look, please do, otherwise I'm happy with the changes.

test/test_hyperparameters.py
```diff
@@ -239,6 +239,18 @@ def test_normalfloat(self):
...
+        with pytest.raises(ValueError):
+            f6 = NormalFloatHyperparameter("param", 5, 10, lower=0.1, upper=0.1, default_value=5.0, q=0.1, log=True)
+
+        f6 = NormalFloatHyperparameter("param", 5, 10, lower=0.1, upper=10, default_value=5.0, q=0.1, log=True)
+        f6_ = NormalFloatHyperparameter("param", 5, 10, lower=0.1, upper=10, default_value=5.0, q=0.1, log=True)
+        self.assertEqual(f6, f6_)
+        self.assertEqual(
+            "param, Type: NormalFloat, Mu: 5.0 Sigma: 10.0, Range: [0.1, 10.0], Default: 5.0, on log-scale, Q: 0.1", str(f6))
+
+        self.assertNotEqual(f1, f2)
```
mfeurer, line 251: This is new duplicated.

mfeurer (changes requested):
Hey, thanks a lot for the PR. I just also had a look and think there are a few things missing:
* test for sampling from the truncated normal distribution
* you need to adapt the get_neighbors functions to take the bounds into account as well
* please also see my inline comments

ConfigSpace/hyperparameters.pyx
```diff
@@ -740,9 +802,19 @@ cdef class NormalFloatHyperparameter(FloatHyperparameter):
...
+            sigma = self.sigma
+            return rs.normal(mu, sigma, size=size)
+        else:
+            mu = self.mu
+            sigma = self.sigma
+            lower = self.lower
+            upper = self.upper
+            a = (self.lower - mu) / sigma
+            b = (upper - mu) / sigma
+            return truncnorm.rvs(a, b, loc=mu, scale=sigma, size=size, random_state=rs)
```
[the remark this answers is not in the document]
Deathn0t: This scaling is mentioned in the [Scipy documentation for truncated normal distribution](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.truncnorm.html)
in short the following formula has to be applied:
`a, b = (myclip_a - my_mean) / my_std, (myclip_b - my_mean) / my_std`
In our experiments, it worked fine like that.

ConfigSpace/hyperparameters.pyx
```diff
@@ -655,11 +662,55 @@ cdef class NormalFloatHyperparameter(FloatHyperparameter):
         self.default_value = self.check_default(default_value)
         self.normalized_default_value = self._inverse_transform(self.default_value)
 
+        if lower is not None and upper is not None:
```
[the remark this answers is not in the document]
Deathn0t: I added this check
```
        if (lower is not None) ^ (upper is not None):
            raise ValueError("Only one bound was provided when both lower and upper bounds must be provided.")
```

with corresponding tests for each sub case

ConfigSpace/hyperparameters.pyx
```diff
@@ -700,16 +753,25 @@ cdef class NormalFloatHyperparameter(FloatHyperparameter):
             sigma=self.sigma,
             log=self.log,
             q=self.q,
+            lower=self.lower,
```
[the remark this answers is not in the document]
Deathn0t: Yes, I added this test to verify this case
```
f1 = NormalFloatHyperparameter("param", 0.5, 10.5)
        f1_ = NormalFloatHyperparameter("param", 0.5, 10.5)
        self.assertEqual(f1, f1_)
        self.assertEqual(
            "param, Type: NormalFloat, Mu: 0.5 Sigma: 10.5, Default: 0.5",
            str(f1))

        # Test attributes are accessible
        self.assertEqual(f1.name, "param")
        self.assertAlmostEqual(f1.mu, 0.5)
        self.assertAlmostEqual(f1.sigma, 10.5)
        self.assertAlmostEqual(f1.q, None)
        self.assertEqual(f1.log, False)
        self.assertAlmostEqual(f1.default_value, 0.5)
        self.assertAlmostEqual(f1.normalized_default_value, 0.5)
```

ConfigSpace/hyperparameters.pyx
```diff
@@ -1153,6 +1255,8 @@ cdef class NormalIntegerHyperparameter(IntegerHyperparameter):
...
+            upper=self.upper,
```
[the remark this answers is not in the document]
Deathn0t: Yes, I followed the same code as NormalFloat with the `round(int(...))`
```
    def to_uniform(self, z: int = 3) -> 'UniformIntegerHyperparameter':
        if self.lower is None or self.upper is None:
            lb = np.round(int(self.mu - (z * self.sigma)))
            ub = np.round(int(self.mu + (z * self.sigma)))
        else:
            lb = self.lower
            ub = self.upper

        return UniformIntegerHyperparameter(self.name,
                                            lb,
                                            ub,
                                            default_value=self.default_value,
                                            q=self.q, log=self.log)
```

test/test_hyperparameters.py
Deathn0t: I am sorry, I don't see the duplicated? I followed the previous test cases with `f4, f4_` which was already there.

Deathn0t:
I added this test for the sampling:

```
    def test_sample_NormalFloatHyperparameter_with_bounds(self):
        hp = NormalFloatHyperparameter("nfhp", 0, 1, lower=-3, upper=3)

        def actual_test():
            rs = np.random.RandomState(1)
            counts_per_bin = [0 for i in range(11)]
            for i in range(100000):
                value = hp.sample(rs)
                index = min(max(int((np.round(value + 0.5)) + 5), 0), 9)
                counts_per_bin[index] += 1

            self.assertEqual([0, 0, 0, 2184, 13752, 34078, 34139, 13669,
                              2178, 0, 0], counts_per_bin)

            self.assertIsInstance(value, float)
            return counts_per_bin

        self.assertEqual(actual_test(), actual_test())
```

Deathn0t:
Hello, I think I answered all the comments. While editing the `get_neighbors` function I found it quite inconsistent... especially for when comparing to other types of hyperparameters. In the case of normal hyperparameters having a "neighbor" based on the variance of the distribution is strange no? The neighbor can be a value out of the bounds given by the function `to_uniform` for example.

Deathn0t:
Hello, any update on this @eddiebergman @mfeurer?

eddiebergman:
> Hello, any update on this @eddiebergman @mfeurer?

Hi @Deathn0t,

Sorry for the delay and thanks for pinging us again on it. I reviewed the PR and I am happy with the changes. @mfeurer will be available later this week or next and I don't want to merge without a final look over by him. Apologies again for this delay. 

In the meantime I dug a little deeper and there are some discussion points if you'd like to have a future say in how we handle some inconsistencies you brought to our attention. 

> While editing the `get_neighbors` function I found it quite inconsistent... especially when comparing to other types of hyperparameters. In the case of normal hyperparameters having a "neighbor" based on the variance of the distribution is strange no? The neighbor can be a value out of the bounds returned by the function `to_uniform` for example.
> 
> I was not able to find a documentation of the expected behaviour for this function.

* `get_neighbors` - Seems to `n` parameter values that are _neighbors_ to the current value, by some definition of _neighbor_ or _close_.
    * This function for `NormalFloatHypereparameter` as it stands has an odd definition of _neighbors_, it just creates a normal distribution centered on whatever `value` is passed and then draws `n` _neighbors_ from that distribution. 
    `neighbours = [ rs.normal(value, self.sigma) for _ in range(n) ]`.
    * I'm not sure of my own definition but to me it would make more sense to have this sample from some reduced `sigma`, i.e.
     `neighbours = [ rs.normal(value, self.sigma / 3) for _ in range(n) ]`, where 3 is arbitrary and a more reasonable value should be used.
    * What would your definition of a _neighbor_ be?
    * The bounding method you use creates a non-normal distribution. For example, consider the case where `lower` and `upper` are close to `value` (mean) but where `self.sigma` is quite wide. This bounding would result in a large amount of _neighbors_ with the same value as `lower` or `upper`. I don't really know how to tackle this to be honest. Would users still expect neighbors to be normally distributed within these bounds or be aware of the fact that with tight bounds, most of the neighbors will be on the boundary.
    ```
                new_value = rs.normal(value, self.sigma)
                if self.lower is not None and self.upper is not None:
                    new_value = min(max(new_value, self.lower), self.upper)
    ``` 
* `to_uniform` - Well, converts a normally distributed parameter to a uniform one.
    * You're right, `NormalFloatHyperparameter.get_neighbors` could return a value outside of the `UniformFloatHypereparameter` that is created by `to_uniform`. While for our own use cases, I don't think this is an issue (maybe?) I could see how this might cause unexpected behaviors.
    * I'm not sure there's a fully sound solution to this but please correct me and suggest if I am wrong here. 
        * A normal distribution is technically unbounded in the values it could return. (without it being specified by params as you did)
        * A uniform distribution is bounded (or at least in how we have it implemeneted)
        * There is no _fully safe_ way to specify a boundary on the normal distribution and so how can it be set for the uniform distribution.
        * Without a formally correct way then in practice we have to pick some value. I believe `mu +- 3*sigma` is reasonable but if you have other suggestions, do let us know.

PR #191, opened 2021-09-01
Extend configuration space to act as mapping

eddiebergman:
This PR extends `ConfigurationSpace` and `Configuration` classes to implement the [Mapping interface ](https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes). This makes them behave more like dictionaries, extending their functionality to allow `config.items(), config.values()` as well as features such as unpacking `mydict = {**config}`.

This PR also removes an old flake8 package which did not seem to do anything and which has now been replaced with native functionality in flake8.

Lastly, it renames a file that tested an `irace` writer which no longer seems to exist. This was renamed to that `pytest` will skip this file during its running.

There are 3 tests added, 2 to test the functionality described for `Configuration` and `ConfigurationSpace` and 1 to ensure the order of hyperparemeters for a Config matches those of the ConfigSpace, as advertised in the doc string.


ConfigSpace/configuration_space.pyx
```diff
@@ -744,6 +747,9 @@ class ConfigurationSpace(object):
         """
         return list(self._hyperparameters.keys())
 
+    def __getitem__(self, key: str) -> Hyperparameter:
+        return self._hyperparameters[key]
```
mfeurer, line 751: What do you think about calling `get_hyperparameter` here?

test/test_configuration_space.py
```diff
@@ -771,6 +771,39 @@ def test_subspace_switches(self):
...
+        indexing and items, values, keys.
+        """
+        cs = ConfigurationSpace()
+        names = [f"name{i}" for i in range(5)]
+        hyperparameters = [
+            UniformIntegerHyperparameter(name, 0, 10)
+            for name in names
+        ]
+        for hp in hyperparameters:
+            cs.add_hyperparameter(hp)
```
mfeurer, line 786: you could use `cs.add_hyperparameters(hyperparameters)`

ConfigSpace/configuration_space.pyx
eddiebergman: Makes sense, has the validation checks

PR #195, opened 2021-09-15
Check new neighbors deterministically

### LINKED ISSUE #194: One-exchange neighborhood may raise `ValueError` non-deterministically
# Description
`ConfigSpace.util.get_one_exchange_neighbourhood` uses `np.random` to decide whether a new configuration is to be validated (see [util.pyx:218](https://github.com/automl/ConfigSpace/blob/9fd85c4e7278a471311bd7172e1910bbbe2e850a/ConfigSpace/util.pyx#L218)). Using `np.random` for this decision makes the behavior of the function non-deterministic.

# Solution proposal
Use the randomness generator `random` instantiated at [util.pyx:130](https://github.com/automl/ConfigSpace/blob/9fd85c4e7278a471311bd7172e1910bbbe2e850a/ConfigSpace/util.pyx#L130) instead of `np.random`.

filipbartek:
One-exchange neighborhood iterator rigorously checks some of the generated configurations. The check is performed with the probability 5 %, as decided by `np.random`.

Make the check deterministic (using an explicitly seeded `np.random.RandomState`) instead of relying on the default numpy random state, which need not be seeded.

Resolves #194.

### DIFF
--- a/ConfigSpace/util.pyx
+++ b/ConfigSpace/util.pyx
@@ -215,7 +215,7 @@ def get_one_exchange_neighbourhood(
                     # Only rigorously check every tenth configuration (
                     # because moving around in the neighborhood should
                     # just work!)
-                    if np.random.random() > 0.95:
+                    if random.random() > 0.95:
                         new_configuration.is_valid_configuration()
                     else:
                         configuration_space._check_forbidden(new_array)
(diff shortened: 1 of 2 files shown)


mfeurer (approved):

mfeurer (changes requested):
Thanks a lot for this fix. Unfortunately, this currently breaks one test which has a fixture that's based on the random seed used (which now changed). Could you please have a look into that?

filipbartek:
> this currently breaks one test which has a fixture that's based on the random seed used (which now changed). Could you please have a look into that?
I have updated the fixture to match the current behavior of the library.

Issue #196, opened 2021-09-17
Invalid configurations generated with constraints


Kerilk:
Hello,

Thanks for this great library. I ran into a small issue while double checking some results. The (not so) small sample here seems to be generating invalid configurations:
```python
import ConfigSpace as CS
import ConfigSpace.hyperparameters as CSH


cs = CS.ConfigurationSpace(seed=1234)
p1 = CSH.CategoricalHyperparameter(name='p1', choices=['None', '#pragma omp #p3', '#pragma omp target #p3', '#pragma omp target #p5', '#pragma omp #p4'])
p3 = CSH.CategoricalHyperparameter(name='p3', choices=['None', '#parallel for #p4', '#parallel for #p6', '#parallel for #p7'])
p4 = CSH.CategoricalHyperparameter(name='p4', choices=['None', 'simd'])
p5 = CSH.CategoricalHyperparameter(name='p5', choices=['None', '#dist_schedule static', '#dist_schedule #p11'])
p6 = CSH.CategoricalHyperparameter(name='p6', choices=['None', '#schedule #p10', '#schedule #p11'])
p7 = CSH.CategoricalHyperparameter(name='p7', choices=['None', '#numthreads #p12'])
p10 = CSH.CategoricalHyperparameter(name='p10', choices=['static', 'dynamic'])
p11 = CSH.OrdinalHyperparameter(name='p11', sequence=['1', '8', '16'])
p12 = CSH.OrdinalHyperparameter(name='p12', sequence=['1', '8', '16'])

cs.add_hyperparameters([p1, p3, p4, p5, p6, p7, p10, p11, p12])

cond0 = CS.EqualsCondition(p3, p1, '#pragma omp #p3')
cond1 = CS.EqualsCondition(p3, p1, '#pragma omp target #p3')
cond2 = CS.EqualsCondition(p5, p1, '#pragma omp target #p5')
cond3 = CS.EqualsCondition(p4, p1, '#pragma omp #p4')


cond4 = CS.EqualsCondition(p4, p3, '#parallel for #p4')
cond5 = CS.EqualsCondition(p6, p3, '#parallel for #p6')
cond6 = CS.EqualsCondition(p7, p3, '#parallel for #p7')

cond7 = CS.EqualsCondition(p11, p5, '#dist_schedule #p11')

cond8 = CS.EqualsCondition(p10, p6, '#schedule #p10')
cond9 = CS.EqualsCondition(p11, p6, '#schedule #p11')

[...]

PR #197, opened 2021-09-17
Generate neighbors and sample configurations correctly with deeply nested conditions

### LINKED ISSUE #194: One-exchange neighborhood may raise `ValueError` non-deterministically

filipbartek:
Resolves #194.

### DIFF
--- a/ConfigSpace/c_util.pyx
+++ b/ConfigSpace/c_util.pyx
@@ -284,25 +284,16 @@ cpdef np.ndarray change_hp_value(
 
     configuration_array[index] = hp_value
 
-    # Hyperparameters which are going to be set to inactive
-    disabled = []
-
     # Activate hyperparameters if their parent node got activated
     children = children_of[hp_name]
     if len(children) > 0:
         to_visit = deque()  # type: deque
         to_visit.extendleft(children)
-        visited = set()  # type: Set[str]
         activated_values = dict()  # type: Dict[str, Union[int, float, str]]
 
         while len(to_visit) > 0:
             current = to_visit.pop()
             current_name = current.name
-            if current_name in visited:
-                continue
-            visited.add(current_name)
-            if current_name in disabled:
-                continue
 
             current_idx = configuration_space._hyperparameter_idx[current_name]
             current_value = configuration_array[current_idx]
@@ -330,19 +321,6 @@ cpdef np.ndarray change_hp_value(
                 children = children_of[current_name]
 
                 if len(children) > 0:
-                    to_disable = set()
-                    for ch in children:
-                        to_disable.add(ch.name)
-                    while len(to_disable) > 0:
-                        child = to_disable.pop()
-                        child_idx = configuration_space._hyperparameter_idx[child]
-                        disabled.append(child_idx)
-                        children = children_of[child]
-
-                        for ch in children:
-                            to_disable.add(ch.name)
-
-    for idx in disabled:
-        configuration_array[idx] = NaN
+                    to_visit.extendleft(children)
 
     return configuration_array
--- a/ConfigSpace/c_util.pyx
+++ b/ConfigSpace/c_util.pyx
@@ -214,29 +214,34 @@ cpdef np.ndarray change_hp_value(
     cdef DTYPE_t NaN = np.NaN
     cdef dict children_of = configuration_space._children_of
 
-    # We maintain to_visit as a minimum heap of indices of hyperparameters that may need to be updated.
-    # We assume that the hyperparameters are sorted topologically by their index.
+    # We maintain `to_visit` as a minimum heap of indices of hyperparameters that may need to be updated.
+    # We assume that the hyperparameters are sorted topologically with respect to the conditions by the hyperparameter indices.
+    # Initially, we know that the hyperparameter with the index `index` may need to be updated (by changing its value to `hp_value`).
     to_visit = [index]
     
-    # Since one hyperparameter may be reachable in several ways, we need to make sure we don't process it twice.
+    # Since one hyperparameter may be reachable in more than one way, we need to make sure we don't schedule it for inspection more than once.
     scheduled = np.zeros(len(configuration_space), dtype=bool)
     scheduled[index] = True
 
-    # Activate hyperparameters if their parent node got activated
+    # Activate hyperparameters if their parent node got activated.
     while len(to_visit) > 0:
         assert np.all(scheduled[to_visit])
         current_idx = heapq.heappop(to_visit)
         current_name = configuration_space._idx_to_hyperparameter[current_idx]
         conditions = configuration_space._parent_conditions_of[current_name]
 
+        # Should the current hyperparameter be active?
         active = True
         for condition in conditions:
             if not condition._evaluate_vector(configuration_array):
+                # The current hyperparameter should be inactive because `condition` is not satisfied. 
                 active = False
                 break
 
+        # Should the value of the current hyperparameter be updated?
         update = False
         if current_idx == index:
+            # The current hyperparameter should be updated because the caller requested this update.
             if not active:
                 raise ValueError(
                     "Attempting to change the value of the inactive hyperparameter '%s' to '%s'." % (hp_name, hp_value))
@@ -245,10 +250,12 @@ cpdef np.ndarray change_hp_value(
         else:
             current_value = configuration_array[current_idx]
             if active and not current_value == current_value:
+                # The current hyperparameter should be active but is inactive.
                 current = configuration_space._hyperparameters[current_name]
                 target_value = current.normalized_default_value
                 update = True
             elif not active and current_value == current_value:
+                # The current hyperparameter should be inactive but is active.
                 # If the hyperparameter was made inactive,
                 # all its children need to be deactivated as well
                 target_value = NaN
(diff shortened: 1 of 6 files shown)


filipbartek:
This solution resolves the constraints in an arbitrary order, possibly updating the status of a HP more than once. My guess is that following a topological order of the HPs (which seems to be respected by `ConfigSpace.get_hyperparameters()` and `ConfigSpace._children_of`) would lead to a linear complexity, while the worst-case complexity of this "arbitrary order" approach is quadratic. Using a topological order would require iterating through all the HPs in the CS, or maintaining `to_visit` as a priority queue.

filipbartek:
I have extended the changes so that generating neighbors and sampling random configurations works correctly with configuration spaces with conditions nested deeper than 1 level of `AbstractConjunction`. Both of these procedures now use the topological order of hyperparameters when ensuring that the output configuration is valid.

ConfigSpace/configuration_space.pyx
```diff
@@ -1269,6 +1269,7 @@ class ConfigurationSpace(collections.abc.Mapping):
 
         unconditional_hyperparameters = self.get_all_unconditional_hyperparameters()
         hyperparameters_with_children = list()
+        conditional_hyperparameters = sorted(self.get_all_conditional_hyperparameters(), key=lambda hp_name: self._hyperparameter_idx[hp_name])
```
filipbartek, line 1272: n ... number of conditional hyperparameters
N ... number of all hyperparameters

This implementation has the complexity O(n log n d(N)), where d(N) is the complexity of fetching an element from a dictionary indexed by N strings.

Alternative:
`[hp_name for hp_name in self._hyperparameters if hp_name in self.get_all_conditional_hyperparameters()]`
Complexity: O(N s(n)), where s(n) is the complexity of checking membership in a set of n strings.

ConfigSpace/c_util.pyx
```diff
@@ -266,83 +204,64 @@ cpdef np.ndarray change_hp_value(
...
-    configuration_array[index] = hp_value
-
-    # Hyperparameters which are going to be set to inactive
-    disabled = []
+    # We maintain to_visit as a minimum heap of indices of hyperparameters that may need to be updated.
+    # We assume that the hyperparameters are sorted topologically by their index.
+    to_visit = [index]
+    
+    # Since one hyperparameter may be reachable in several ways, we need to make sure we don't process it twice.
+    scheduled = np.zeros(len(configuration_space), dtype=bool)
```
filipbartek, line 222: Does using a `np.ndarray` without declaring it with `cdef` harm performance significantly?

filipbartek:
This pull request probably solves the issue solved by #219.

mfeurer:
Hi @dengdifan and @filipbartek,
Please excuse the delay, but I finally get to your two PRs (this one and #219). As a first step I checked, and they both fix the issue in https://github.com/automl/ConfigSpace/pull/219/files#diff-62b93f19eeb2c5ff6e26eaccf3f63ee6641e87f3a6ccfc2ffb0dec86925ae245

As a next step I'd like to figure why the two PRs are so different in size and whether they implement different things underneath.

From what I can see the PR of @difandeng touches the functions:
* `c_util.change_hp_value`

and the PR from @filipbartek touches

* `c_util.correct_sampled_array`
* `c_util.change_hp_value`
* `configuration_space.ConfigSpace.get_active_hyperparameters`

and none of your PRs touches
* `c_util.check_configurations()`
* `util.deactivate_inactive_hyperparameters()`

PR #194 is more complete than PR #219 as it handles the test added there but not vice versa.

I do like the rather small changes of @difandeng to `c_util.change_hp_value` and from @filipbartek to `configuration_space.ConfigSpace.get_active_hyperparameters`. However, I don't immediately understand the large changes from @filipbartek to `c_util.correct_sampled_array` and `c_util.change_hp_value`, could you please give some further explanation on what you are doing here? Would it be possible to reproduce the changed behavior in the style of the original algorithms? Lastly, we'd need to check the two functions that weren't touched by any of the PRs to ensure that they don't require an update.

Therefore, I propose the following concrete steps forward, let me know what you think about them:
1. Merge #219
2. @filipbartek rebases this PR on master and keeps the changes from #219 for `c_util.change_hp_value` and changes `c_util.correct_sampled_array` to be closer to the original version of the algorithm. We then merge this PR, too
3. We check whether `c_util.check_configurations()` and `util.deactivate_inactive_hyperparameters()` require any special treatment, too.

mfeurer:
We need to check if this PR fixes https://github.com/automl/ConfigSpace/issues/196 as well.

dengdifan:
This should also fix https://github.com/automl/ConfigSpace/issues/253

ConfigSpace/c_util.pyx
```diff
@@ -121,110 +121,48 @@ cpdef np.ndarray correct_sampled_array(
...
-            if child_name not in inactive:
-                parents = parents_of[child_name]
-                hyperparameter_idx = hyperparameter_to_idx[child_name]
-                if len(parents) == 1:
-                    conditions = parent_conditions_of[child_name]
-                    add = True
-                    for j in range(len(conditions)):
-                        condition = conditions[j]
-                        if not condition._evaluate_vector(vector):
-                            add = False
```
filipbartek, line 186: This may be too early to declare `child` inactive. We know at this point that `child` has 1 parent and that at least one of its parent conditions is not satisfied. However, if `conditions` are in disjunction, the `child` still has a chance of being active.

ConfigSpace/c_util.pyx
filipbartek, line 222: `scheduled` is a replacement of `visited`. While `visited` is a set of HP names, `scheduled` is a boolean vector representation of a set of HPs, where each HP is identified by its index. I estimate that this change is probably a premature potential optimization.

ConfigSpace/c_util.pyx
```diff
@@ -266,83 +204,64 @@ cpdef np.ndarray change_hp_value(
...
+            assert current_idx > index
             current_value = configuration_array[current_idx]
-
-            conditions = configuration_space._parent_conditions_of[current_name]
-
-            active = True
-            for condition in conditions:
-                if not condition._evaluate_vector(configuration_array):
-                    active = False
-                    break
```
filipbartek, line 316: The value of `active` for the HP `current` need not be definitive. For example, let's assume that `current` is being deactivated because one of its parent conditions does not hold. Since we do not process the HPs in topological order, there may be a future iteration of the `to_visit` loop that activates one of the parents of `current`. If `current` is conditioned by an `OrConjunction`, `current` may need to be re-activated then. However, we will not get to re-visit and re-activate `current` afterward because of its membership in `visited`.
Processing the HPs in topological order resolves this issue.

ConfigSpace/c_util.pyx
```diff
@@ -266,83 +204,64 @@ cpdef np.ndarray change_hp_value(
...
+        for condition in conditions:
+            if not condition._evaluate_vector(configuration_array):
+                active = False
+                break
+
+        update = False
+        if current_idx == index:
+            if not active:
+                raise ValueError(
+                    "Attempting to change the value of the inactive hyperparameter '%s' to '%s'." % (hp_name, hp_value))
```
filipbartek, line 241: As a nice side effect of the code reorganization, we explicitly prevent the caller from changing the value of an inactive HP.

filipbartek:
> [...] I don't immediately understand the large changes from @filipbartek to `c_util.correct_sampled_array` and `c_util.change_hp_value`, could you please give some further explanation on what you are doing here? Would it be possible to reproduce the changed behavior in the style of the original algorithms? [...]

To explain the changes in `change_hp_value`, I have extended the code comments with commit [3b0dbf1](https://github.com/automl/ConfigSpace/pull/197/commits/3b0dbf144fa2ceee23e6857e9b1aa42fa8973fea) and added some commit comments here on GitHub.
When I modified the function, I attempted to make the modification rather conservative, keeping namely variable names and semantics where possible. However, I could still stay closer to the original implementation. I can try to undo some of the changes (namely switching from set `visited` to bitvector `scheduled`, dropping unused variable `activated_values`, preventing code duplication by introducing `update`) to concentrate the patch on the crucial stuff. Would you prefer that, @mfeurer?

I agree with merging #219 first.

I will try to document the changes in `correct_sampled_array` and respond to the other concerns you raised later.

mfeurer:
Hi @filipbartek, thank you very much for your explanation. As discussed, I just merged #219. Could you please rebase your PR, I will then check again how to proceed.

filipbartek:
I am trying to minimize the changes introduced by this PR. 3 functions are affected:
- `c_util.correct_sampled_array`
- `c_util.change_hp_value`
- `configuration_space.ConfigSpace.get_active_hyperparameters`
I believe each of these is broken in a distinct way and can be fixed independently. Breaking the fix down might further facilitate the inspection of the changes. @mfeurer, would you prefer to have 3 PRs for these 3 fixes, or does breaking them down into sets of commits suffice?

eddiebergman:
Sorry to close this after so long of inactivity. I did a major rework in #346 and the code around this has completely changed. It's also been benchmarked to be just as fast/faster than the existing solution. I'm not sure if this closes the existing issue #194 though, I will have to take a deeper look

filipbartek:
Thanks for the update, Eddie!
> I'm not sure if this closes the existing issue #194 though, I will have to take a deeper look
Consider scavenging the tests from this pull request.

PR #203, opened 2021-12-08
add check for pcs_new read

dengdifan:
convert the corresponding forbidden values correctly to the target hyperparameters type

### DIFF
--- a/ConfigSpace/read_and_write/pcs_new.py
+++ b/ConfigSpace/read_and_write/pcs_new.py
@@ -414,11 +414,33 @@ def read(pcs_string, debug=False):
             else:
                 # So far, only equals is supported by SMAC
                 if tmp_list[1] == '=':
-                    # TODO maybe add a check if the hyperparameter is
-                    # actually in the configuration space
+                    hp = configuration_space.get_hyperparameter(tmp_list[0])
+                    if isinstance(hp, NumericalHyperparameter):
+                        if isinstance(hp, IntegerHyperparameter):
+                            forbidden_value = int(tmp_list[2])
+                        elif isinstance(hp, FloatHyperparameter):
+                            forbidden_value = float(tmp_list[2])
+                        if forbidden_value < hp.lower or forbidden_value > hp.upper:
+                            # TODO alternatively we could raise an warning and ignore this forbidden
+                            raise ValueError(f'forbidden_value is set out of the bound, it needs to be set between'
+                                             f'[{hp.lower}, {hp.upper}] but its value is {forbidden_value}')
+                    elif isinstance(hp, (CategoricalHyperparameter, OrdinalHyperparameter)):
+                        hp_values = hp.choices if isinstance(hp, CategoricalHyperparameter) else hp.sequence
+                        forbidden_value_in_hp_values = False
+                        for hp_value in hp_values:
+                            if str(hp_value) == tmp_list[2]:
+                                forbidden_value = hp_value
+                                forbidden_value_in_hp_values = True
+                                break
+                        if not forbidden_value_in_hp_values:
+                            raise ValueError(f'forbidden_value is set out of the bound, it needs to be set as'
+                                             f'one member of {hp_values}  but its value is {forbidden_value}')
+                    else:
+                        raise ValueError('Unsupported Hyperparamter sorts')
+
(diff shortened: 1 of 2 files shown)


test/read_and_write/test_pcs_converter.py
```diff
@@ -398,6 +398,40 @@ def test_build_new_GreaterThanIntCondition(self):
...
+    def test_read_new_configuration_space_forbidden(self):
+        cs_with_forbidden = ConfigurationSpace()
+        int_hp = UniformIntegerHyperparameter('int_hp', 0, 50, 30)
+        float_hp = UniformFloatHyperparameter('float_hp', 0., 50., 30.)
+        cat_hp_str = CategoricalHyperparameter('cat_hp_str', ['a', 'b', 'c'], 'b')
+        ord_hp_str = OrdinalHyperparameter('ord_hp_str', ['a', 'b', 'c'], 'b')
+
+        cs_with_forbidden.add_hyperparameters([int_hp, float_hp, cat_hp_str, ord_hp_str])
+
+        int_hp_forbidden = ForbiddenAndConjunction(ForbiddenEqualsClause(int_hp, 1))
```
mfeurer, line 410: Not sure if this really needs to be wrapped inside a `ForbiddenAndConjunction`.

test/read_and_write/test_pcs_converter.py
```diff
@@ -398,6 +398,40 @@ def test_build_new_GreaterThanIntCondition(self):
...
+
+        float_hp_forbidden_1 = ForbiddenEqualsClause(float_hp, 1.0)
+        float_hp_forbidden_2 = ForbiddenEqualsClause(float_hp, 2.0)
+        float_hp_forbidden = ForbiddenAndConjunction(float_hp_forbidden_1, float_hp_forbidden_2)
+
+        cat_hp_str_forbidden = ForbiddenAndConjunction(ForbiddenEqualsClause(cat_hp_str, 'a'))
```
mfeurer, line 416: Same here

test/read_and_write/test_pcs_converter.py
```diff
@@ -398,6 +398,40 @@ def test_build_new_GreaterThanIntCondition(self):
...
+        ord_hp_float_forbidden = ForbiddenAndConjunction(ForbiddenEqualsClause(ord_hp_str, 'a'))
```
mfeurer, line 417: And here

mfeurer: I'm actually surprised. We do allow for one value passed to a `Conjunction`?

test/read_and_write/test_pcs_converter.py
dengdifan: Otherwise, the two generated ConfigSpace does not equal. the pcs_reader generates the forbidden_clauses with the following code:
https://github.com/dengdifan/ConfigSpace/blob/pcs_read/ConfigSpace/read_and_write/pcs_new.py#L450

ConfigSpace/read_and_write/pcs_new.py
```diff
@@ -414,11 +414,36 @@ def read(pcs_string, debug=False):
...
+                            # TODO alternatively we could raise an warning and ignore this
+                            raise ValueError(f'forbidden_value is set out of the bound, it needs to'
+                                             f' be set between [{hp.lower}, {hp.upper}]'
+                                             f' but its value is {forbidden_value}')
+                    elif isinstance(hp, (CategoricalHyperparameter, OrdinalHyperparameter)):
+                        hp_values = hp.choices if isinstance(hp, CategoricalHyperparameter)\
+                            else hp.sequence
+                        forbidden_value_in_hp_values = False
+                        for hp_value in hp_values:
+                            if str(hp_value) == tmp_list[2]:
```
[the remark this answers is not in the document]
dengdifan: When I implemented this code, I was thinking that hp_value could be a float or int. But since pcs_new reader only transform choices/ sequences to string, this line is redundant, I will fix that

test/read_and_write/test_pcs_converter.py
mfeurer: Thanks for the explanation.

mfeurer (approved):
This looks good, thank you. Do you think such casting needs to be applied for the condition statements as well?

PR #210, opened 2021-12-14
Make print(config) friendlier for copy-paste

mfeurer:
Implements #66.

Configurations are now printed as
```
Configuration(values={
  'chp': '1',
  'const': 1,
  'nfhp': 0.0,
  'nihp': 0,
  'ohp': '1',
  'ufhp': 5.5,
  'uihp': 6,
})
```
opposed to
```
Configuration:
  chp, Value: '1'
  const, Constant: 1
  nfhp, Value: 0.0
  nihp, Value: 0
  ohp, Value: '1'
  ufhp, Value: 5.5
  uihp, Value: 6
```
.

This does not yet allow for copy-pasting the configuration as the argument `configuration_space` is not printed, but it is much closer. Yet another alternative would be
```
{
  'chp': '1',
  'const': 1,
  'nfhp': 0.0,
  'nihp': 0,
  'ohp': '1',
  'ufhp': 5.5,
  'uihp': 6,
}
```
which could be used as the `values` argument to `Configuration`.

If there are any other suggestions I would be happy to hear about them.

CC @eddiebergman @KEggensperger @dengdifan @renesass @mlindauer @RaghuSpaceRajan

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -17,6 +17,10 @@ jobs:
       run: |
         cd docs
         make html
+    - name: Run doctests
+      run: |
+        cd docs
+        make doctest
     - name: Pull latest gh-pages
       if: (contains(github.ref, 'master')) && github.event_name == 'push'
       run: |
--- a/ConfigSpace/configuration_space.pyx
+++ b/ConfigSpace/configuration_space.pyx
@@ -1573,7 +1573,7 @@ class Configuration(collections.abc.Mapping):
         self._populate_values()
 
         representation = io.StringIO()
-        representation.write("Configuration:\n")
+        representation.write("Configuration(values={\n")
 
         hyperparameters = self.configuration_space.get_hyperparameters()
         hyperparameters.sort(key=lambda t: t.name)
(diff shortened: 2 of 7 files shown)


RaghuSpaceRajan:
Looks good, generally. How come you did not add the argument `configuration_space`? Because we do not know the variable name for the argument `configuration_space`?

eddiebergman:
I'm okay with either the current implementation or the pure dict one. In general, as long as `print(dict(config))` gives a pure dict output, then this implementation as you have it is fine with me.

eddiebergman (approved):

KEggensperger:
Yes, this is much better than before. I am fine with the current version, but I would prefer something that is actually valid code, i.e. a pure dict, over something that just looks like its valid but is then missing an argument.

renesass:
I like the dictionary output.

mlindauer:
my gut feeling is that most of the community is looking for `dict`-like input and output. Therefore, I would also prefer this way of printing.

dengdifan:
The current implementation looks good to me. We need to be aware that the printed object is a configuration, not a dict.

mfeurer:
Thanks everyone for your feedback. As we need to balance readability and pastability here I think the current proposal strikes a nice middle ground:
* It keeps the information that we have a Configuration
* by calling `print(dict(config))` one gets the dictionary representation
* it can be extended to include the ConfigSpace argument to which we don't know the variable name at print time

I just also updated the docs and enabled doctests.

PR #211, opened 2021-12-14
Update wheel builder to include windows and osx

eddiebergman:
This PR builds wheels for the same distributions as supported in `sklearn` [github actions](https://github.com/scikit-learn/scikit-learn/blob/main/.github/workflows/wheels.yml).

This action also runs our suite of tests on these wheels which seems to complete without error.

The output of the test can be seen here: 
* https://github.com/eddiebergman/ConfigSpace/runs/4523610401?check_suite_focus=true

This also adds the `workflow_dispatch` event so that it can be triggered manually.

I would still advocate for some manual testing of the wheels produced in the artifacts on a seperate windows/mac device, seperate from the environment it was built on, if possile.

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -1,8 +1,8 @@
 # Workflow to build and test wheels
 # =================================
 # This github action gets triggered whenever there
-# is a push to the master branch or a release is created. 
-# It generates both wheels and distributions files, 
+# is a push to the master branch or a release is created.
+# It generates both wheels and distributions files,
 # making sure their contents are correct via unit-testing.
 #
 # However, only in the case of a github release, the assets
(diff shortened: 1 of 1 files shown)


mfeurer (commented):
That looks good. I'm wondering whether we now also need to run unit tests etc on Windows. We now know that we can build the wheels, but we don't know whether our packages work there.

eddiebergman:
Click in to the tests, it looks like it runs the unittests on those wheels.

mfeurer:
Thanks a lot for the pointer. This is excellent. I'm wondering whether we should test these other platforms during the regular unit tests as well? But that would be a separate PR I assume?

eddiebergman:
Let's do it as a separate PR, if we run into build issues when doing unit-tests then we might want to have dedicated changes there.

PR #217, opened 2022-01-12
Betaparams

hvarfner:
### Added the beta parameter types, the accompanying imports and checks, and implemented the infrastructure for computing the pdf of a parameter

__hyperparameters__: 
- added BetaFloat and BetaIntegerHyperparameter - they share the neighbor generation of their uniform counterparts. 
- Added pdf and _pdf for all parameters, where the logic is
pdf(vector) = _pdf(inverse_transform(vector)). This means that the user can call pdf(X) on the original parameter range, but the acquisition function calls _pdf, since it already recieves the transformed parameter ranges.
- The Beta parameters are not normalized to 0-1 range, which is simply because Normal was not, either. Not sure if this is the correct choice.
- The local searches are just copy-paste of the corresponding Uniform parameter types - with the exception that the jump from existing configuration is scaled by the range of the parameter (as it it not in 0-1, the jump presumably needs to reflect this)

__json__: added the necessary stuff for the new parameters to fit in with the rest
__configuration_space__: Added a method to_uniform to create a ConfigSpace of only uniform parameters - needed for the new local search approach in SMAC (see ei_optimization)

### DIFF
--- a/ConfigSpace/configuration_space.pyx
+++ b/ConfigSpace/configuration_space.pyx
@@ -41,6 +41,9 @@ from ConfigSpace.hyperparameters import (
     Hyperparameter,
     Constant,
     FloatHyperparameter,
+    UniformFloatHyperparameter,
+    UniformIntegerHyperparameter,
+    OrdinalHyperparameter
 )
 from ConfigSpace.conditions import (
     ConditionComponent,
@@ -1341,6 +1344,19 @@ class ConfigurationSpace(collections.abc.Mapping):
         """
         self.random = np.random.RandomState(seed)
 
+    def to_uniform(self) -> 'ConfigurationSpace':
+        uniform_config_space = ConfigurationSpace()
+        for parameter in self.get_hyperparameters():
+            if type(parameter) not in [UniformFloatHyperparameter, UniformIntegerHyperparameter, OrdinalHyperparameter]:
+                uniform_config_space.add_hyperparameter(parameter.to_uniform())
+            else:
+                uniform_config_space.add_hyperparameter(parameter)
+
+        # TODO - check that this is certainly correct    
+        uniform_config_space.add_conditions(self.get_conditions())
+        uniform_config_space.add_forbidden_clauses(self.get_forbiddens())
+        
+        return uniform_config_space
 
 class Configuration(collections.abc.Mapping):
     def __init__(self, configuration_space: ConfigurationSpace,
--- a/ConfigSpace/hyperparameters.pyx
+++ b/ConfigSpace/hyperparameters.pyx
@@ -869,37 +869,16 @@ cdef class NormalFloatHyperparameter(FloatHyperparameter):
             vector = np.log(vector)
         return vector
 
+    def get_neighbors(self, value: float, rs: np.random.RandomState, number: int = 4,
+                      transform: bool = False) -> List[float]:
+        neighbors = []
+        for i in range(number):
+            new_value = rs.normal(value, self.sigma)
 
-    def get_neighbors(
-        self,
-        value: Union[int, float],
-        rs: np.random.RandomState,
-        number: int = 4,
-        transform: bool = False,
-    ) -> List[Union[np.ndarray, float, int]]:
-        neighbors = []  # type: List[Union[np.ndarray, float, int]]
-        while len(neighbors) < number:
-            rejected = True
-            iteration = 0
-            while rejected:
-                iteration += 1
-                new_value = rs.normal(value, self.sigma)
-                int_value = self._transform(value)
-                new_int_value = self._transform(new_value)
+            if self.lower is not None and self.upper is not None:
+                    new_value = min(max(new_value, self.lower), self.upper)
 
-                if self.lower is not None and self.upper is not None:
-                    int_value = min(max(int_value, self.lower), self.upper)
-                    new_int_value = min(max(new_int_value, self.lower), self.upper)
-
-                if int_value != new_int_value:
-                    rejected = False
-                elif iteration > 100000:
-                    raise ValueError('Probably caught in an infinite loop.')
-
-            if transform:
-                neighbors.append(self._transform(new_value))
-            else:
-                neighbors.append(new_value)
+            neighbors.append(new_value)
         return neighbors
 
 
--- a/ConfigSpace/hyperparameters.pyx
+++ b/ConfigSpace/hyperparameters.pyx
@@ -811,7 +811,7 @@ cdef class NormalFloatHyperparameter(FloatHyperparameter):
         else:
             raise ValueError("Illegal default value %s" % str(default_value))
 
-    def to_integer(self) -> NormalIntegerHyperparameter:
+    def to_integer(self) -> 'NormalIntegerHyperparameter':
         if self.q is None:
             q_int = None
         else:
@@ -1058,7 +1058,7 @@ cdef class BetaFloatHyperparameter(FloatHyperparameter):
     def __hash__(self):
         return hash((self.name, self.alpha, self.beta, self.log, self.q))
 
-    def to_uniform(self) -> UniformFloatHyperparameter:
+    def to_uniform(self) -> 'UniformFloatHyperparameter':
         lb = self.lower
         ub = self.upper
 
@@ -1088,7 +1088,7 @@ cdef class BetaFloatHyperparameter(FloatHyperparameter):
         else:
             raise ValueError("Illegal default value %s" % str(default_value))
 
-    def to_integer(self) -> BetaIntegerHyperparameter:
+    def to_integer(self) -> 'BetaIntegerHyperparameter':
         if self.q is None:
             q_int = None
         else:
@@ -2152,7 +2152,7 @@ cdef class CategoricalHyperparameter(Hyperparameter):
 
     # TODO review the name for this - it's just for producing a categorical hyperparameter
     # with equal weights for all choices
-    def to_uniform(self):
+    def to_uniform(self) -> 'CategoricalHyperparameter':
         return CategoricalHyperparameter(
             name=self.name,
             choices=copy.deepcopy(self.choices),
--- a/ConfigSpace/hyperparameters.pyx
+++ b/ConfigSpace/hyperparameters.pyx
@@ -35,7 +35,7 @@ from collections import OrderedDict, Counter
 from typing import List, Any, Dict, Union, Set, Tuple, Optional
 
 import numpy as np
-from scipy.stats import truncnorm, beta, norm, betabinom
+from scipy.stats import truncnorm, beta as spbeta, norm, betabinom
 cimport numpy as np
 
 
@@ -603,6 +603,10 @@ cdef class UniformFloatHyperparameter(FloatHyperparameter):
                 neighbors.append(neighbor)
         return neighbors
 
+    def pdf(self, vector: np.ndarray) -> np.ndarray:
+        ub = self._upper
+        lb = self._lower
+        return np.ones_like(vector) / (ub - lb)
 
 # TODO - implement a .pdf method here, and make sure the log case is properly considerec
 cdef class NormalFloatHyperparameter(FloatHyperparameter):
@@ -789,7 +793,7 @@ cdef class NormalFloatHyperparameter(FloatHyperparameter):
         else:
             raise ValueError("Illegal default value %s" % str(default_value))
 
-    def to_integer(self) -> 'NormalIntegerHyperparameter':
+    def to_integer(self) -> NormalIntegerHyperparameter:
         if self.q is None:
             q_int = None
         else:
@@ -818,6 +822,7 @@ cdef class NormalFloatHyperparameter(FloatHyperparameter):
             upper = self.upper
             a = (self.lower - mu) / sigma
             b = (upper - mu) / sigma
+            
             return truncnorm.rvs(a, b, loc=mu, scale=sigma, size=size, random_state=rs)
 
     cpdef np.ndarray _transform_vector(self, np.ndarray vector):
@@ -869,6 +874,284 @@ cdef class NormalFloatHyperparameter(FloatHyperparameter):
             a = (lower - mu) / sigma
             b = (upper - mu) / sigma
             return truncnorm(a, b, loc=mu, scale=sigma).pdf(vector)
+
+
+
+# and this one if for defining custom discrete distributions over some space 
+# with the combination of the two, you can design most reasonable probability distributions
+# TODO - implement
+cdef class BetaFloatHyperparameter(FloatHyperparameter):
+    cdef public alpha
+    cdef public beta
+
+    def __init__(self, name: str, alpha: Union[int, float], beta: Union[int, float],
+                 default_value: Union[None, float] = None,
+                 q: Union[int, float, None] = None, log: bool = False,
+                 lower: Optional[Union[float, int]] = None,
+                 upper: Optional[Union[float, int]] = None,
+                 meta: Optional[Dict] = None) -> None:
+        r"""
+        A float hyperparameter.
+
+        Its values are sampled from a beta distribution
+        :math:`Beta(\alpha, \beta)`.
+
+        Example
+        -------
+
+        >>> import ConfigSpace as CS
+        >>> import ConfigSpace.hyperparameters as CSH
+        >>> cs = CS.ConfigurationSpace(seed=1)
+        >>> normal_float_hp = CSH.BetaFloatHyperparameter('beta_float', alpha=3,
+        ...                                                 beta=1, log=False)
+        >>> cs.add_hyperparameter(beta_float_hp)
+        beta_float, Type: BetaFloat, Alpha: 3.0 beta: 1.0, Default: 0.5
+
+        Parameters
+        ----------
+        name : str
+            Name of the hyperparameter, with which it can be accessed
+        alpha : int, float
+            Alpha parameter of the normalized beta distribution
+        beta : int, float
+            Beta parameter of the normalized beta distribution
+        default_value : int, float, optional
+            Sets the default value of a hyperparameter to a given value
+        q : int, float, optional
+            Quantization factor
+        log : bool, optional
+            If ``True``, the values of the hyperparameter will be sampled
+            on a logarithmic scale. Default to ``False``
+        lower : int, float, optional
+            Lower bound of a range of values from which the hyperparameter will be sampled.
+            The Beta disribution gets scaled by the total range of the hyperparameter.
+        upper : int, float, optional
+            Upper bound of a range of values from which the hyperparameter will be sampled.
+            The Beta disribution gets scaled by the total range of the hyperparameter.
+        meta : Dict, optional
+            Field for holding meta data provided by the user.
+            Not used by the configuration space.
+        """
+        super(BetaFloatHyperparameter, self).__init__(name, default_value, meta)
+        self.alpha = float(alpha)
+        self.beta = float(beta)
+        self.q = float(q) if q is not None else None
+        self.log = bool(log)
+        self.default_value = self.check_default(default_value)
+        self.normalized_default_value = self._inverse_transform(self.default_value)
+
+        if (lower is not None) ^ (upper is not None):
+            raise ValueError("Only one bound was provided when both lower and upper bounds must be provided.")
+
+        if lower is not None and upper is not None:
+            self.lower = float(lower)
+            self.upper = float(upper)
+
+            if self.lower >= self.upper:
+                raise ValueError("Upper bound %f must be larger than lower bound "
+                                "%f for hyperparameter %s" %
+                                (self.upper, self.lower, name))
+            elif log and self.lower <= 0:
+                raise ValueError("Negative lower bound (%f) for log-scale "
+                                "hyperparameter %s is forbidden." %
+                                (self.lower, name))
+
+            self.default_value = self.check_default(default_value)
+
+            if self.log:
+                if self.q is not None:
+                    lower = self.lower - (np.float64(self.q) / 2. - 0.0001)
+                    upper = self.upper + (np.float64(self.q) / 2. - 0.0001)
+                else:
+                    lower = self.lower
+                    upper = self.upper
+                self._lower = np.log(lower)
+                self._upper = np.log(upper)
+            else:
+                if self.q is not None:
+                    self._lower = self.lower - (self.q / 2. - 0.0001)
+                    self._upper = self.upper + (self.q / 2. - 0.0001)
+                else:
+                    self._lower = self.lower
+                    self._upper = self.upper
+            if self.q is not None:
+                # There can be weird rounding errors, so we compare the result against self.q, see
+                # In [13]: 2.4 % 0.2
+                # Out[13]: 0.1999999999999998
+                if np.round((self.upper - self.lower) % self.q, 10) not in (0, self.q):
+                    raise ValueError(
+                        'Upper bound (%f) - lower bound (%f) must be a multiple of q (%f)'
+                        % (self.upper, self.lower, self.q)
+                    )
+
+    def __repr__(self) -> str:
+        repr_str = io.StringIO()
+        repr_str.write("%s, Type: BetaFloat, Alpha: %s Beta: %s, Range: [%s, %s], Default: %s" % (self.name, repr(self.alpha), repr(self.beta), repr(self.lower), repr(self.upper), repr(self.default_value)))
+
+        if self.log:
+            repr_str.write(", on log-scale")
+        if self.q is not None:
+            repr_str.write(", Q: %s" % str(self.q))
+        repr_str.seek(0)
+        return repr_str.getvalue()
+
+    def __eq__(self, other: Any) -> bool:
+        """
+        This method implements a comparison between self and another
+        object.
+
+        Additionally, it defines the __ne__() as stated in the
+        documentation from python:
+            By default, object implements __eq__() by using is, returning NotImplemented
+            in the case of a false comparison: True if x is y else NotImplemented.
+            For __ne__(), by default it delegates to __eq__() and inverts the result
+            unless it is NotImplemented.
+
+        """
+        if not isinstance(other, self.__class__):
+            return False
+
+        return (
+            self.name == other.name and
+            self.default_value == other.default_value and
+            self.alpha == other.alpha and
+            self.beta == other.beta and
+            self.log == other.log and
+            self.q == other.q and
+            self.lower == other.lower and
+            self.upper == other.upper
+        )
+
+    def __copy__(self):
+        return BetaFloatHyperparameter(
+            name=self.name,
+            default_value=self.default_value,
+            alpha=self.alpha,
+            beta=self.beta,
+            log=self.log,
+            q=self.q,
+            lower=self.lower,
+            upper=self.upper,
+            meta=self.meta
+        )
+
+    def __hash__(self):
+        return hash((self.name, self.alpha, self.beta, self.log, self.q))
+
+    def to_uniform(self) -> UniformFloatHyperparameter:
+        lb = self.lower
+        ub = self.upper
+
+        return UniformFloatHyperparameter(self.name,
+                                          lb,
+                                          ub,
+                                          default_value=self.default_value,
+                                          q=self.q, log=self.log)
+
+    def check_default(self, default_value: Union[int, float]) -> Union[int, float]:
+        # return mode as default
+        lb = self.lower
+        ub = self.upper
+        
+        if default_value is None:
+            if (self.alpha > 1) and (self.beta > 1):
+                return (self.alpha - 1) / (self.alpha + self.beta - 2)
+            elif self.alpha > self.beta:
+                return lb
+            elif self.alpha > self.beta:
+                return ub
+            else: 
+                return (ub - lb) / 2
+
+        elif self.is_legal(default_value):
+            return default_value
+        else:
+            raise ValueError("Illegal default value %s" % str(default_value))
+
+    def to_integer(self) -> BetaIntegerHyperparameter:
+        if self.q is None:
+            q_int = None
+        else:
+            q_int = int(self.q)
+        return BetaIntegerHyperparameter(self.name, int(self.mu), self.sigma,
+                                           default_value=int(np.round(self.default_value, 0)),
+                                           q=q_int, log=self.log)
+
+    def is_legal(self, value: Union[float]) -> bool:
+        return isinstance(value, float) or isinstance(value, int)
+
+    cpdef bint is_legal_vector(self, DTYPE_t value):
+        return isinstance(value, float) or isinstance(value, int)
+
+    def _sample(self, rs: np.random.RandomState, size: Optional[int] = None
+                ) -> Union[np.ndarray, float]:
+
+        if self.lower == None:
+            return ValueError('The beta hyperparameter must have both bounds specified.')
+        else:
+            alpha = self.alpha
+            beta = self.beta
+            lower = self.lower
+            upper = self.upper
+            return spbeta(alpha, beta, loc=lower, scale=upper-lower).rvs(size=size, random_state=rs)
+
+    cpdef np.ndarray _transform_vector(self, np.ndarray vector):
+        if np.isnan(vector).any():
+            raise ValueError('Vector %s contains NaN\'s' % vector)
+        if self.log:
+            vector = np.exp(vector)
+        if self.q is not None:
+            vector = np.rint(vector / self.q) * self.q
+        return vector
+
+    cpdef double _transform_scalar(self, double scalar):
+        if scalar != scalar:
+            raise ValueError('Number %s is NaN' % scalar)
+        if self.log:
+            scalar = math.exp(scalar)
+        if self.q is not None:
+            scalar = round(scalar / self.q) * self.q
+        return scalar
+
+    def _inverse_transform(self, vector: Optional[np.ndarray]) -> Union[float, np.ndarray]:
+        if vector is None:
+            return np.NaN
+
+        if self.log:
+            vector = np.log(vector)
+        return vector
+
+    def get_neighbors(
+        self,
+        value: Any,
+        rs: np.random.RandomState,
+        number: int = 4,
+        transform: bool = False,
+        std: float = 0.2
+    ) -> List[float]:
+        neighbors = []  # type: List[float]
+        while len(neighbors) < number:
+            neighbor = rs.normal(value, std)  # type: float
+            if neighbor < 0 or neighbor > 1:
+                continue
+            if transform:
+                neighbors.append(self._transform(neighbor))
+            else:
+                neighbors.append(neighbor)
+        return neighbors
+
+    def pdf(self, vector: np.ndarray) -> np.ndarray:
+        alpha = self.alpha
+        beta = self.beta
+        if self.lower == None:
+            return ValueError('The beta hyperparameter must have both bounds specified.')
+        else:
+            lower = self.lower
+            upper = self.upper
+            return spbeta(alpha, beta, loc=lower, scale=upper-lower).pdf(vector)
+
+    
+
 
 
 # TODO - implement a .pdf method here, and make sure the log case is properly considered
@@ -1376,7 +1659,7 @@ cdef class NormalIntegerHyperparameter(IntegerHyperparameter):
 # and this one if for defining custom discrete distributions over some space 
 # with the combination of the two, you can design most reasonable probability distributions
 # TODO - implement
-cdef class BetaBinomialIntegerHyperparameter(NumericalHyperparameter):
+cdef class BetaIntegerHyperparameter(NumericalHyperparameter):
     def __init__(self, name: str, default_value: int, meta: Optional[Dict] = None) -> None:
         super(IntegerHyperparameter, self).__init__(name, default_value, meta)
         pass
--- a/ConfigSpace/hyperparameters.pyx
+++ b/ConfigSpace/hyperparameters.pyx
@@ -2235,10 +2235,13 @@ cdef class CategoricalHyperparameter(Hyperparameter):
         except ValueError:
             return None
 
-    def _inverse_transform(self, vector: Union[None, str, float, int]) -> Union[int, float]:
+
+    # TODO - encountered something of a bug here when this was passed through .pdf
+    # for some reason, it wanted a numpy array as output
+    def _inverse_transform(self, vector: Union[None, np.ndarray, str, float, int]) -> Union[np.ndarray, int, float]:
         if vector is None:
             return np.NaN
-        return self.choices.index(vector)
+        return np.array(self.choices.index(vector))
 
     def has_neighbors(self) -> bool:
         return len(self.choices) > 1
@@ -2289,6 +2292,12 @@ cdef class CategoricalHyperparameter(Hyperparameter):
                          "OrdinalHyperparameter, but is "
                          "<cdef class 'ConfigSpace.hyperparameters.CategoricalHyperparameter'>")
 
+    def _pdf(self, vector: np.ndarray) -> np.ndarray:
+        return np.array(self.probabilities[vector])
+
+    def pdf(self, vector: np.ndarray) -> np.ndarray:
+        vector = self._inverse_transform(vector)
+        return self._pdf(vector)
 
 
 # TODO - implement .pdf method
@@ -2577,3 +2586,8 @@ cdef class OrdinalHyperparameter(Hyperparameter):
     def allow_greater_less_comparison(self) -> bool:
         return True
 
+    def _pdf(self, vector: np.ndarray) -> np.ndarray:
+        return self.ones_like(vector) / self.num_elements
+
+    def pdf(self, vector: np.ndarray) -> np.ndarray:
+        return self._pdf(vector) 
--- a/ConfigSpace/hyperparameters.pyx
+++ b/ConfigSpace/hyperparameters.pyx
@@ -898,9 +898,6 @@ cdef class NormalFloatHyperparameter(FloatHyperparameter):
             return truncnorm(a, b, loc=mu, scale=sigma).pdf(vector)
 
 
-# and this one if for defining custom discrete distributions over some space 
-# with the combination of the two, you can design most reasonable probability distributions
-# TODO - implement
 cdef class BetaFloatHyperparameter(FloatHyperparameter):
     cdef public alpha
     cdef public beta
@@ -1600,7 +1597,6 @@ cdef class NormalIntegerHyperparameter(IntegerHyperparameter):
             meta=self.meta
         )
 
-    # todo check if conversion should be done in initiation call or inside class itsel
     def to_uniform(self, z: int = 3) -> 'UniformIntegerHyperparameter':
         if self.lower is None or self.upper is None:
             lb = np.round(int(self.mu - (z * self.sigma)))
@@ -1705,9 +1701,6 @@ cdef class NormalIntegerHyperparameter(IntegerHyperparameter):
         return self.nfhp._pdf(vector) / self.normalization_constant
 
 
-# and this one if for defining custom discrete distributions over some space 
-# with the combination of the two, you can design most reasonable probability distributions
-# TODO - implement
 cdef class BetaIntegerHyperparameter(IntegerHyperparameter):
     cdef public alpha
     cdef public beta
@@ -1865,7 +1858,6 @@ cdef class BetaIntegerHyperparameter(IntegerHyperparameter):
             meta=self.meta
         )
 
-    # todo check if conversion should be done in initiation call or inside class itsel
     def to_uniform(self, z: int = 3) -> 'UniformIntegerHyperparameter':
         lb = self.lower
         ub = self.upper
@@ -2016,7 +2008,6 @@ cdef class BetaIntegerHyperparameter(IntegerHyperparameter):
         return self.bfhp._pdf(vector) / self.normalization_constant
 
 
-# TODO - implement .pdf method
 cdef class CategoricalHyperparameter(Hyperparameter):
     cdef public tuple choices
     cdef public tuple probabilities
@@ -2143,8 +2134,8 @@ cdef class CategoricalHyperparameter(Hyperparameter):
             meta=self.meta
         )
 
-    # TODO review the name for this - it's just for producing a categorical hyperparameter
-    # with equal weights for all choices
+    # This is just for the uniform configspace for PiBO
+    # It creates a categorical parameter with equal weights for all choices
     def to_uniform(self) -> 'CategoricalHyperparameter':
         return CategoricalHyperparameter(
             name=self.name,
@@ -2238,8 +2229,8 @@ cdef class CategoricalHyperparameter(Hyperparameter):
             return None
 
 
-    # TODO - encountered something of a bug here when this was passed through .pdf
-    # for some reason, it wanted a numpy array as output
+    # TEncountered something of a bug here when this was passed through .pdf
+    # for some reason, it wanted a numpy array as output (and as such, it is now changed)
     def _inverse_transform(self, vector: Union[None, np.ndarray, str, float, int]) -> Union[np.ndarray, int, float]:
         if vector is None:
             return np.NaN
@@ -2302,7 +2293,6 @@ cdef class CategoricalHyperparameter(Hyperparameter):
         return self._pdf(vector)
 
 
-# TODO - implement .pdf method
 cdef class OrdinalHyperparameter(Hyperparameter):
     cdef public tuple sequence
     cdef public int num_elements
--- a/ConfigSpace/hyperparameters.pyx
+++ b/ConfigSpace/hyperparameters.pyx
@@ -379,8 +379,9 @@ cdef class FloatHyperparameter(NumericalHyperparameter):
         raise NotImplementedError()
 
     def pdf(self, vector):
-        raise NotImplementedError()
-    
+        vector = self._inverse_transform(vector)
+        return self._pdf(vector)
+
 
 cdef class IntegerHyperparameter(NumericalHyperparameter):
     def __init__(self, name: str, default_value: int, meta: Optional[Dict] = None) -> None:
@@ -417,9 +418,10 @@ cdef class IntegerHyperparameter(NumericalHyperparameter):
 
     cpdef np.ndarray _transform_vector(self, np.ndarray vector):
         raise NotImplementedError()
-
+    
     def pdf(self, vector):
-        raise NotImplementedError()
+        vector = self._inverse_transform(vector)
+        return self._pdf(vector)
 
 
 cdef class UniformFloatHyperparameter(FloatHyperparameter):
@@ -618,7 +620,7 @@ cdef class UniformFloatHyperparameter(FloatHyperparameter):
                 neighbors.append(neighbor)
         return neighbors
 
-    def pdf(self, vector: np.ndarray) -> np.ndarray:
+    def _pdf(self, vector: np.ndarray) -> np.ndarray:
         ub = self._upper
         lb = self._lower
         return np.ones_like(vector) / (ub - lb)
@@ -838,7 +840,7 @@ cdef class NormalFloatHyperparameter(FloatHyperparameter):
             upper = self._upper
             a = (lower - mu) / sigma
             b = (upper - mu) / sigma
-            
+
             return truncnorm.rvs(a, b, loc=mu, scale=sigma, size=size, random_state=rs)
 
     cpdef np.ndarray _transform_vector(self, np.ndarray vector):
@@ -879,7 +881,7 @@ cdef class NormalFloatHyperparameter(FloatHyperparameter):
             neighbors.append(new_value)
         return neighbors
 
-    def pdf(self, vector: np.ndarray) -> np.ndarray:
+    def _pdf(self, vector: np.ndarray) -> np.ndarray:
         mu = self.mu
         sigma = self.sigma
         if self.lower == None:
@@ -889,7 +891,10 @@ cdef class NormalFloatHyperparameter(FloatHyperparameter):
             # as the data always comes in transformed
             lower = self._lower
             upper = self._upper
-            return truncnorm(lower, upper, loc=mu, scale=sigma).pdf(vector)
+            a = (lower - mu) / sigma
+            b = (upper - mu) / sigma
+                
+            return truncnorm(a, b, loc=mu, scale=sigma).pdf(vector)
 
 
 # and this one if for defining custom discrete distributions over some space 
@@ -1106,8 +1111,8 @@ cdef class BetaFloatHyperparameter(FloatHyperparameter):
         else:
             alpha = self.alpha
             beta = self.beta
-            lower = self.lower
-            upper = self.upper
+            lower = self._lower
+            upper = self._upper
             return spbeta(alpha, beta, loc=lower, scale=upper-lower).rvs(size=size, random_state=rs)
 
     cpdef np.ndarray _transform_vector(self, np.ndarray vector):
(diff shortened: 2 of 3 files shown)


ConfigSpace/configuration_space.pyx
```diff
@@ -1341,6 +1344,19 @@ class ConfigurationSpace(collections.abc.Mapping):
...
+            if type(parameter) not in [UniformFloatHyperparameter, UniformIntegerHyperparameter, OrdinalHyperparameter]:
+                uniform_config_space.add_hyperparameter(parameter.to_uniform())
+            else:
+                uniform_config_space.add_hyperparameter(parameter)
+
+        # TODO - check that this is certainly correct    
+        uniform_config_space.add_conditions(self.get_conditions())
+        uniform_config_space.add_forbidden_clauses(self.get_forbiddens())
+        
+        return uniform_config_space
```
eddiebergman, line 1359: I have to check this later, you add the uniform version of parameters but do you require the non_uniform parameters to be removed from the config space?

ConfigSpace/hyperparameters.pyx
```diff
@@ -339,6 +345,9 @@ cdef class NumericalHyperparameter(Hyperparameter):
             meta=self.meta
         )
 
+    def pdf(self, vector):
```
eddiebergman, line 348: Add `vector` type and return type

ConfigSpace/hyperparameters.pyx
```diff
@@ -151,6 +151,9 @@ cdef class Hyperparameter(object):
     cpdef int compare_vector(self, DTYPE_t value, DTYPE_t value2):
         raise NotImplementedError()
 
+    def pdf(self, value):
```
eddiebergman, line 154: Add return type

ConfigSpace/hyperparameters.pyx
```diff
@@ -369,6 +378,10 @@ cdef class FloatHyperparameter(NumericalHyperparameter):
     cpdef np.ndarray _transform_vector(self, np.ndarray vector):
         raise NotImplementedError()
 
+    def pdf(self, vector):
```
eddiebergman, line 381: same

ConfigSpace/hyperparameters.pyx
```diff
@@ -405,6 +418,10 @@ cdef class IntegerHyperparameter(NumericalHyperparameter):
 
     cpdef np.ndarray _transform_vector(self, np.ndarray vector):
         raise NotImplementedError()
+    
+    def pdf(self, vector):
```
eddiebergman, line 422: same

ConfigSpace/hyperparameters.pyx
```diff
@@ -858,6 +882,303 @@ cdef class NormalFloatHyperparameter(FloatHyperparameter):
...
 
 
+    def _pdf(self, vector: np.ndarray) -> np.ndarray:
+        mu = self.mu
+        sigma = self.sigma
+        if self.lower == None:
+            raise ValueError('Need upper and lower limits when using user priors.')
+        else:
+            # regardless of whether we have log input or not, this yields the correct answer,
+            # as the data always comes in transformed
```
eddiebergman, line 892: Commenet above seems to conflict the todo note at the top of the class. 

Either:
* remove the TODO if this comment is correct
* check for the `log` bool being set and return a `NotImplementedError`

ConfigSpace/hyperparameters.pyx
```diff
@@ -858,6 +882,303 @@ cdef class NormalFloatHyperparameter(FloatHyperparameter):
...
+            self.default_value == other.default_value and
+            self.alpha == other.alpha and
+            self.beta == other.beta and
+            self.log == other.log and
+            self.q == other.q and
+            self.lower == other.lower and
+            self.upper == other.upper
+        )
+
+    def __copy__(self):
```
eddiebergman, line 1042: I would prefer types on magic methods, add a return type

ConfigSpace/hyperparameters.pyx
```diff
@@ -858,6 +882,303 @@ cdef class NormalFloatHyperparameter(FloatHyperparameter):
...
+            alpha=self.alpha,
+            beta=self.beta,
+            log=self.log,
+            q=self.q,
+            lower=self.lower,
+            upper=self.upper,
+            meta=self.meta
+        )
+
+    def __hash__(self):
```
eddiebergman, line 1055: same

eddiebergman (commented):
I will have a proper look over the logic later but this all requires tests, including the new `_pdf` functionality. I know this is required for multi-objective `SMAC` to be released but seeing as `SMAC` and has quite a libraries rely on this, I wouldn't feel good about merging until it's tested.

ConfigSpace/configuration_space.pyx
hvarfner: Sure, this is a bit unintuitive. Let's see if I understood you correctly:

I'm generating a ConfigurationSpace of only uniform parameters (or categoricals without weights in that specific case). However, the original one is also kept, and both are fed into SMAC. The reason for this is that the local search associated with PiBO samples a fraction of its starting points uniformly, and the other half from the prior (which has proven empirically superior to only doing one or the other). As such, we need two ConfigurationSpaces.

Let me know if I didn't answer your question!

ConfigSpace/hyperparameters.pyx
hvarfner: Yep, miss on my end. Same for rest.

ConfigSpace/hyperparameters.pyx
hvarfner: Thanks, great catch. It is indeed correct, and the to-do note will be removed.

ConfigSpace/hyperparameters.pyx
hvarfner: Sure!

hvarfner: Wait, sorry. None of them had return types as of now, so it's be an inconsistency for me to add them to the Beta parameters. Are you sure I should add a return type?

hvarfner:
Thanks @eddiebergman . I can write some number of tests for the _pdf and pdf methods that I think are sensible. However, it would be great to possibly have a discussion about what you expect from this.

ConfigSpace/configuration_space.pyx
eddiebergman: Sorry, I missed the very first line of the function call. This seems correct except for your TODO which is something to test.

I think `config_space.to_uniform()` is not very intuitive of it's functionality.
What does it mean for a `config_space` to be _uniform_? While this interpretation makes sense, there could be other interpretations that make sense.

Considering it's one of the main API points for external uses, it could do with some documentation on that point.

If it's also a major API point, it could do with a little example with explanation in [the docs](https://github.com/hvarfner/ConfigSpace/blob/betaparams/docs/source/User-Guide.rst) (which are sorely in need of some updating).

hvarfner: Ah, that makes sense. Yeah, I think a more precise name is a good idea. I could definitely do some docs writing, as well.

I think something like remove_parameter_priors or make_parameters_uniform is ultimately a more descriptive name for the method.

eddiebergman:
Seeing as `pdf` is essentially a new feature to `ConfigSpace`, it needs to be documented and tested, otherwise it doesn't exist as far as anyone else using the library would know about.

I pulled your fork to try and get a nice overview of the `pdf` function and whats expected from it. I would document in the `Hyperparameter` class what calling `pdf` should return to make things easier. I guess it's a probability density function but how is that represented through code? 

For classes that rely on a subclass having a private method that is called, this should be documented, for example in `FloatHyperparameter`, it's not currently obvious that any class subclassing it would need to implement `_pdf`:
```python
class FloatHyperparameter(NumericalHyperparameter):
    
    def _pdf(self, vector: np.ndarray) -> np.ndarray:
        """
        something about _pdf needing to be implemented and what it is being passes        
        """
        raise NotImplementedError()

    
    def pdf(self, vector: np.ndarray) -> np.ndarray:
        vector = self._inverse_transform(vector)
        return self._pdf(vector)

``` 

The same applies to `IntegerHyperparameter`

Once the documentation for that is done, it should indicate what tests are required. I actually have no clue what you would want to test, or how to test it but whatever you document that the `pdf` function does, it should do it for all `Hyperparameter` types that support it. For those that don't, the test should instead expected a `NotImplementedError`, meaning if it were to be implemented in the future, then we will know that it now needs to be tested properly.

As a side note, the fact this is hard to identify is an indication we should use `ABC` to indicate abstract classes and `@abstractmethod` to indicate methods that should be overwritten to make it easy to contribute. I'm not sure how well this integrates with `.pyx` files that get compiled.

eddiebergman:
>   Nevermind the failed test, a few lines in my latest merge ended up in the wrong class.

>   Hi Eddie,
    I'm working on what I hope are the last bits and pieces of ConfigSpace, and the accompanying SMAC changes are presumably done. I have made sure to do some proper documetation, and I am now in the midst of writing the unit tests (one of them seems to fail in master, the get_size for normal int). I assume I should get to work on doing complete testing for the Beta parameters, and pdf's for all the parameters. Right?
    Also, is there anything else that you see needs to be done, apart from what was mentioned in the PR?
    Best Regards,
    Carl

Hi @hvarfner,

I would suggest keeping conversation here so I will see it more frequently and allows for others such as @mfeurer to add to the discussion if required.

Yes, having test for the new function you introduced is the most important part, it will likely illuminate some bugs as it seems to have been found in your most recent commits.

As far as I can tell, these are the only change I would require. Seeing as commit d04eee9 seems to introduce another method, this should be documented too.

Best,
Eddie

hvarfner:
Hi @eddiebergman ,

Sorry for being a bit quiet. Sounds good! I've been working pretty frenetically on this over the last few days, and I have indeed found some bugs that are being sorted out. My best guess is that another PR will appear by the end of the day!

Best Regards,
Carl

hvarfner:
@eddiebergman I updated the main text as well, hopefully that helps with sorting out everything.

hvarfner:
@eddiebergman Unifortunately found one bug in the sampling/default values of logged Normal parameters. Working to sort it out atm.

eddiebergman:
No problem, im in the middle of reviewing the current version, taking a while just due to the size

ConfigSpace/hyperparameters.pyx
```diff
@@ -877,6 +998,327 @@ cdef class NormalFloatHyperparameter(FloatHyperparameter):
...
+        else:
+            q_int = int(self.q)
+        return BetaIntegerHyperparameter(self.name, lower=self.lower, upper=self.upper, alpha=self.alpha, beta=self.beta,
+                                           default_value=int(np.round(self.default_value, 0)),
+                                           q=q_int, log=self.log)
+
+    def is_legal(self, value: Union[float]) -> bool:
+        if not (isinstance(value, float) or isinstance(value, int)):
+            return False
+        elif self.upper >= value >= self.lower:
```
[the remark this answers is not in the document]
eddiebergman: Might add more complication around transforming the value if it's been constructed with `log=True`.

ConfigSpace/hyperparameters.pyx
```diff
@@ -877,6 +998,327 @@ cdef class NormalFloatHyperparameter(FloatHyperparameter):
...
+            return True
+        else:
+            return False
+
+    cpdef bint is_legal_vector(self, DTYPE_t value):
+        if self.upper >= value >= self.lower:
```
eddiebergman, line 1228: Likewise here?

ConfigSpace/hyperparameters.pyx
```diff
@@ -877,6 +998,327 @@ cdef class NormalFloatHyperparameter(FloatHyperparameter):
...
+            return True
+        else:
+            return False
+
+    def _sample(self, rs: np.random.RandomState, size: Optional[int] = None
+                ) -> Union[np.ndarray, float]:
+
+        if (self.lower == None) or (self.upper == None):
+            return ValueError('The beta hyperparameter must have both bounds specified.')
```
eddiebergman, line 1237: If the parameter is made non-optional, then this check is not needed.

ConfigSpace/hyperparameters.pyx
```diff
@@ -877,6 +998,327 @@ cdef class NormalFloatHyperparameter(FloatHyperparameter):
...
+                neighbors.append(self._transform(neighbor))
+            else:
+                neighbors.append(neighbor)
+        return neighbors
+
+    def _pdf(self, vector: np.ndarray) -> np.ndarray:
+        alpha = self.alpha
+        beta = self.beta
+        if self.lower == None:
+            return ValueError('The beta hyperparameter must have both bounds specified.')
```
eddiebergman, line 1295: Same here for error checking if it's non-optional

ConfigSpace/hyperparameters.pyx
```diff
@@ -877,6 +998,327 @@ cdef class NormalFloatHyperparameter(FloatHyperparameter):
...
+            upper = self._upper
+            return spbeta(alpha, beta, loc=lower, scale=upper-lower).pdf(vector)
+
+    def get_max_density(self) -> float:
+        lb = self.lower
+        ub = self.upper
+
+        if (self.alpha > 1) and (self.beta > 1):
+            normalized_mode = (self.alpha - 1) / (self.alpha + self.beta - 2)
+            return self._pdf(np.array([(self._upper - self._lower) * normalized_mode + self._lower]))[0]
```
eddiebergman, line 1307: You use the `_lower` and `_upper`, the transformed variants of the bounds, but in the next statements use `self.lower` and `self.upper`, the non-transformed versions. Just want to make sure this is intentional.

ConfigSpace/hyperparameters.pyx
```diff
@@ -1395,10 +1863,324 @@ cdef class NormalIntegerHyperparameter(IntegerHyperparameter):
...
+            Sets the default value of a hyperparameter to a given value
+        q : int, optional
+            Quantization factor
+        log : bool, optional
+            If ``True``, the values of the hyperparameter will be sampled
+            on a logarithmic scale. Defaults to ``False``
+        lower : int, float, optional
+            Lower bound of a range of values from which the hyperparameter will be sampled
+        upper : int, float, optional
+            Upper bound of a range of values from which the hyperparameter will be sampled
```
eddiebergman, line 1913: Same here with making these non-optional and looking through the class to remove error checking on it.

ConfigSpace/hyperparameters.pyx
```diff
@@ -1483,7 +2265,8 @@ cdef class CategoricalHyperparameter(Hyperparameter):
         repr_str.write("}")
         repr_str.write(", Default: ")
         repr_str.write(str(self.default_value))
-        if self.probabilities is not None:
+        # if the probability distribution is not uniform, write out the probabilities
+        if not np.all(self.probabilities == self.probabilities[0]):
```
eddiebergman, line 2269: If `weights` is `None` during construction, this will fail as `self.probabilties` is `None`

eddiebergman: Never mind, this was fixed

ConfigSpace/hyperparameters.pyx
```diff
@@ -1659,6 +2455,44 @@ cdef class CategoricalHyperparameter(Hyperparameter):
...
+        distributions, only normal distributions (as the inverse_transform
+         in the pdf method handles these).
+        vector: np.ndarray
+            the (N, ) vector of imputs for which the probability density
+            function is to be computed.
+        """
+        probs = np.array(self.probabilities)
+        res = np.array(probs[vector])
+        if res.ndim == 0:
+            return res.reshape(-1)
```
eddiebergman, line 2490: How do you get a 0 dimensional numpy array and what does it mean to reshape it? Don't think I've ever encountered one. If this is the case, can you just return an explicit object, i.e. `np.array([])` or whatever it's meant to be

eddiebergman (changes requested):
Seems a lot more cohesive now :) There are still some issues and minor comments. I didn't check all the tests since they follow mostly the same pattern but since you are finding issues with it and it seems pretty comprehensive, I will trust you on it.

### Just as a note to myself and other developers, not requested changes
It seems like most classes could use properties to avoid issues of knowing when and how to use each `self.lower` or `self._lower`. This could probabaly be extended to quite a few different properties of the objects to make things simpler:

```python
@property
def lower(self) -> Union[int, float]:
    # ..compute a normalized/log transformed lower if needed 
```

The typing of some functions could definitely be improved as well
```python
from numbers import Real
def f_accepts_float_and_int(n: Real) -> Real:
    ...
```

The `hyperparameters.pyx` file is getting way too large as well as the `test_hyperparameters.py` file, these should definitely be split up.

docs/source/User-Guide.rst
```diff
@@ -242,3 +242,43 @@ To read it from file
...
+Consider the case of optimizing the accuracy of an MLP with three hyperparameters: learning rate [1e-5, 1e-1], dropout [0, 0.99] and activation {Tanh, ReLU}. From prior experience, you believe the optimal learning rate to be around 1e-3, a good dropout to be around 0.25, and the optimal activation function to be ReLU about 80% of the time. This can be represented accordingly:
+
+>>> import numpy as np
+>>> import ConfigSpace.hyperparameters as CSH
+>>> from ConfigSpace.configuration_space import ConfigurationSpace
+>>> # convert 10 log to natural log for learning rate, mean 1e-3
+>>> logmean = np.log(np.power(10.0, -3))
+>>> # two standard deviations on either side of the mean to cover the search space
+>>> logstd = np.log(10.0) 
+>>> learning_rate = CSH.NormalFloatHyperparameter(name='learning_rate', lower=1e-5, upper=1e-1, default_value=1e-3, mu=logmean, sigma=logstd, log=True)
```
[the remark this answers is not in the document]
hvarfner: I agree. This is exactly what I'm trying to look at right now. Also, there's some silly stuff going on with the default values (if none is given) in that it assigns
`self.default_value=self.mu` 
which is often going to be out of bounds for the log case (and doesn't correspond to the default value the user had in mind.

What we have now corresponds to when an actual log-normal distribution is defined, but it's not optimal for intuition.

docs/source/User-Guide.rst
```diff
@@ -242,3 +242,43 @@ To read it from file
...
+>>> dropout = CSH.BetaFloatHyperparameter(name='dropout', lower=0, upper=0.99, default_value=0.25, alpha=2, beta=4, log=False)
+>>> activation = CSH.CategoricalHyperparameter(name='activation', choices=['tanh', 'relu'], weights=[0.2, 0.8])
+
+>>> cs = ConfigurationSpace()
+>>> cs.add_hyperparameters([learning_rate, dropout, activation])
+Configuration space object:
+  Hyperparameters:
+    activation, Type: Categorical, Choices: {tanh, relu}, Default: tanh, Probabilities: (0.2, 0.8)
```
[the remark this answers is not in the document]
hvarfner: Sure!

docs/source/User-Guide.rst
```diff
@@ -242,3 +242,43 @@ To read it from file
...
+    dropout, Type: BetaFloat, Alpha: 2.0 Beta: 4.0, Range: [0.0, 0.99], Default: 0.25
+    learning_rate, Type: NormalFloat, Mu: -6.907755278982137 Sigma: 2.302585092994046, Range: [1e-05, 0.1], Default: 0.001, on log-scale
+
+To check that your prior makes sense for each hyperparameter, you can easily do so with the __pdf__ method:
```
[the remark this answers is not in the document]
hvarfner: Will fix!

test/test_configuration_space.py
```diff
@@ -803,6 +804,24 @@ def test_acts_as_mapping(self):
...
+        cs.add_hyperparameter(CategoricalHyperparameter('cat', [0, 1, 2], weights=[1, 2, 3]))
+        cs.add_hyperparameter(BetaFloatHyperparameter("beta", alpha=8, beta=2, lower=-1, upper=11))
+        cs.add_hyperparameter(NormalIntegerHyperparameter("norm", mu=5, sigma=4, lower=1, upper=15))
+
+        uniform_cs = cs.remove_parameter_priors()
+        
+        expected_cs = ConfigurationSpace()
+        expected_cs.add_hyperparameter(UniformIntegerHyperparameter('integer', 1, 5, log=True))
+        expected_cs.add_hyperparameter(CategoricalHyperparameter('cat', [0, 1, 2]))
+        # These default values are computed closed form for beta, trivial for normal
```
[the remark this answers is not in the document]
hvarfner: The default is the mode of the distribution for all distributionss (if nothing else is given as default), so that could be said at the very least! 

I'll add what you proposed.

test/test_hyperparameters.py
```diff
@@ -82,11 +83,96 @@ def test_constant(self):
         # test that meta-data is stored correctly
         c1_meta = Constant("value", 1, dict(self.meta_data))
         self.assertEqual(c1_meta.meta, self.meta_data)
-
         # Test getting the size
         for constant in (c1, c2, c3, c4, c5, c1_meta):
             self.assertEqual(constant.get_size(), 1)
 
+    def test_constant_pdf(self):
```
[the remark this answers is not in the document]
hvarfner: Ah yeah, true!

test/test_hyperparameters.py
```diff
@@ -82,11 +83,96 @@ def test_constant(self):
...
+        self.assertEqual(tuple(c1._pdf(array_2)), tuple(np.array([0.0, 0.0])))
+        self.assertEqual(tuple(c1._pdf(array_3)), tuple(np.array([1.0, 0.0])))
+
+        # pdf must take a numpy array
+        with self.assertRaises(TypeError):
+            c1._pdf(0.2)
+        with self.assertRaises(TypeError):
+            c1._pdf('pdf')
+
+        # Simply check that it runs, since _pdf does not restrict shape (only public method does)
```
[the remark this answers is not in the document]
hvarfner: I'm not sure I follow here! Can you elaborate?

ConfigSpace/hyperparameters.pyx
```diff
@@ -608,6 +718,17 @@ cdef class UniformFloatHyperparameter(FloatHyperparameter):
                 neighbors.append(neighbor)
         return neighbors
 
+    def _pdf(self, vector: np.ndarray) -> np.ndarray:
+        ub = 1
+        lb = 0
```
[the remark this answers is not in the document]
hvarfner: I agree. I'll change it!

ConfigSpace/hyperparameters.pyx
```diff
@@ -877,6 +998,327 @@ cdef class NormalFloatHyperparameter(FloatHyperparameter):
             return np.rint((self.upper - self.lower) / self.q) + 1
 
 
+    def _pdf(self, vector: np.ndarray) -> np.ndarray:
+        mu = self.mu
+        sigma = self.sigma
```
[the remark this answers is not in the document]
hvarfner: WIll fix

ConfigSpace/hyperparameters.pyx
```diff
@@ -877,6 +998,327 @@ cdef class NormalFloatHyperparameter(FloatHyperparameter):
...
+                                          ub,
+                                          default_value=self.default_value,
+                                          q=self.q, log=self.log)
+
+    def check_default(self, default_value: Union[int, float]) -> Union[int, float]:
+        # return mode as default
+        lb = self.lower
+        ub = self.upper
+        
+        if (default_value is None) or (default_value < lb) or (default_value > ub):
```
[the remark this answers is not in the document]
hvarfner: Sure. Now, it's

`if (lower is None) or (upper is None):
            raise ValueError("Both lower and upper bounds must be provided for a beta distribution.")
`
which I guess doesn't make much sense.

docs/source/User-Guide.rst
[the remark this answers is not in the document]
eddiebergman: Okay, that makes sense but yeah as you point out it's not intuitive, I made a note at the bottom of this review and I think `self.default_value` is something that could be wrapper in a property to ensure it's always giving what's intended.

ConfigSpace/hyperparameters.pyx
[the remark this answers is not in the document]
hvarfner: I believe this one is only used on the ready-transformed values. I was a bit lazy here, and made the check identical to the Uniform counterparts, but all I've seen so far is that it only operates on ready-transformed values.

ConfigSpace/hyperparameters.pyx
hvarfner: Same answer =)

ConfigSpace/hyperparameters.pyx
hvarfner: And fixed

ConfigSpace/hyperparameters.pyx
hvarfner: fixed

ConfigSpace/hyperparameters.pyx
hvarfner: If you feed a 1-dimensional vector of one value here, so e.g.
`vector = np.array([1])` 
which has shape (1, ), and you then retrieve the index, `probs[vector]`, you will actually get the correct value, but in a zero-dimensional vector (so equivalent to `np.array(5)` without brackets). This is a zero-dimensional vector. However, this does not happen if the vector is of any other shape than exactly (1, ).

If you reshape it, you get the desired (1, ) shape (so the same shape as the input for this specific case).

test/test_hyperparameters.py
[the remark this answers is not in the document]
eddiebergman: I should have put the review comment slightly lower but here's the code for reference.

```python
# Simply check that it runs, since _pdf does not restrict shape (only public method does)
self.assertEqual(c1._pdf(accepted_shape_1)[0][0], 1.0)
self.assertEqual(c1._pdf(accepted_shape_2)[0][0], 1.0)
self.assertEqual(c1._pdf(accepted_shape_2)[0][2], 0.0)
self.assertEqual(c1._pdf(accepted_shape_3)[0][0], 0.0)
self.assertEqual(c1._pdf(accepted_shape_3)[2][0], 1.0)
```

Sure, suppose someone in the future comes along and makes some changes to `_pdf`, they run the tests and find out these lines above fail. They then might consider "oh I need to ensure these tests pass" and then code in what's required to make sure these tests above will pass.

We now have the case where `_pdf` is made to explicitly pass those tests and produce those outputs given those inputs. This means the tests have defined how `_pdf` should work. From you `#` comment above, it seems you just wanted to make sure it doesn't crash, and not actually say, this is what `_pdf` should return given a 2d input.

There's two solutions I would consider
* One is just to run `c1._pdf(accepted_shape)` without testing for output, this way you're testing that it accepts it but make no condition on what it returns.
* The other solution would be to explicitly prohibit the dimensions that can come into `_pdf` ... this sort of invalidates the checking in `_pdf` and would make me think each class just implements it's own `pdf` function and checking, leading to some slight duplication but a cleaner interface, i.e. less jumping around the code to understand a function.

I'll leave it to you to decide though, I think ConfigSpace will eventually be due for some cleaning up but it's a while away :)

ConfigSpace/configuration_space.pyx
hvarfner: That's true. Will fix!

ConfigSpace/hyperparameters.pyx
eddiebergman: Ahh okay, didn't realise you could do `np.array(5)` and that it gives the output it does. Thanks for the pointer, all's good :)

ConfigSpace/hyperparameters.pyx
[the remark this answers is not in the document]
eddiebergman: Feel free to resolve these once fixed :)

ConfigSpace/hyperparameters.pyx
```diff
@@ -824,10 +944,11 @@ cdef class NormalFloatHyperparameter(FloatHyperparameter):
         else:
             mu = self.mu
             sigma = self.sigma
-            lower = self.lower
-            upper = self.upper
-            a = (self.lower - mu) / sigma
+            lower = self._lower
+            upper = self._upper
+            a = (lower - mu) / sigma
```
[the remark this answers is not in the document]
Deathn0t: Good catch! Yes I missed this quantisation feature.

hvarfner:
@eddiebergman We're getting there now! However, I'm getting error in configspace this when I'm working on a nicer example for SMAC:

`INFO:smac.utils.io.cmd_reader.CMDReader:Output to smac3-output_2022-01-21_15:21:23_234268
INFO:smac.facade.smac_hpo_facade.SMAC4HPO:Optimizing a deterministic scenario for quality without a tuner timeout - will make SMAC deterministi
c and only evaluate one configuration per iteration!
Traceback (most recent call last):
  File "/home/hvarfner/Documents/SMAC3/examples/python/user_prior_mlp.py", line 154, in <module>
    smac = SMAC4HPO(
  File "/home/hvarfner/Documents/SMAC3/smac/facade/smac_hpo_facade.py", line 90, in __init__
    super().__init__(**kwargs)
  File "/home/hvarfner/Documents/SMAC3/smac/facade/smac_ac_facade.py", line 229, in __init__
    scenario.write()
  File "/home/hvarfner/Documents/SMAC3/smac/scenario/scenario.py", line 169, in write
    self.out_writer.write_scenario_file(self)
  File "/home/hvarfner/Documents/SMAC3/smac/utils/io/output_writer.py", line 63, in write_scenario_file
    new_value = self._parse_argument(scenario, key, getattr(scenario, key))
  File "/home/hvarfner/Documents/SMAC3/smac/utils/io/output_writer.py", line 110, in _parse_argument
    self.save_configspace(scenario.cs, pcs_path, 'pcs_new')  # type: ignore[attr-defined] # noqa F821
  File "/home/hvarfner/Documents/SMAC3/smac/utils/io/output_writer.py", line 199, in save_configspace
    fh.write(writer(cs))
  File "/home/hvarfner/Documents/ConfigSpace/ConfigSpace/read_and_write/pcs_new.py", line 591, in write
    param_lines.write(build_categorical(hyperparameter))
  File "/home/hvarfner/Documents/ConfigSpace/ConfigSpace/read_and_write/pcs_new.py", line 93, in build_categorical
    raise ValueError('The pcs format does not support categorical hyperparameters with '
ValueError: The pcs format does not support categorical hyperparameters with assigend weights/probabilities (for hyperparameter activation)`

Do you have any idea what this is? I have two CategoricalHyperparameters:

`    activation = CategoricalHyperparameter(
        'activation', ['logistic', 'tanh', 'relu'], weights=[1, 1, 3])
    # Moreover, we believe ADAM is the most likely optimizer
    optimizer = CategoricalHyperparameter(
        'optimizer', ['sgd', 'adam'], weights=[1, 2])`

docs/source/User-Guide.rst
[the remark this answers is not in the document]
hvarfner: We can definitely remove it, but I unfortunately think this one becomes a bit insensible if we do! I would rather add another parameter in the example, which I can look to do.

ConfigSpace/hyperparameters.pyx
hvarfner: @eddiebergman I was wrong here, but it is now fixed!

docs/source/User-Guide.rst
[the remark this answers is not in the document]
hvarfner: @eddiebergman Okay, so now this is what happens:

You still define the mean and standard deviation in this slightly awkward way. However, I do think that this is best, as you don't have to guess (but just go to Wikipedia) if you don't know your lognormal distribution by heart. If you do not specify the default value as above, it would still have 1e-3 as the default, and say as much.

However, mu and sigma would have the slightly awkward values of -6.9 and 2.3 which, once again, is mathematically correct. An option would be to add some sort of parenthesis to the repr() of the parameter, but this is probably not going to look great.

eddiebergman:
I did a search for the error message through the ConfigSpace code and found it mentioned in two files,  [`pcs.py`](https://github.com/automl/ConfigSpace/blob/master/ConfigSpace/read_and_write/pcs.py) and [`pcs_new.py`](https://github.com/automl/ConfigSpace/blob/master/ConfigSpace/read_and_write/pcs_new.py). They both are near identical and is the source of the errors, apparently it doesn't like writing categoricals with probabilities attached to file.

Then a code search of `pcs` and `pcs_new` through SMAC ended up with a few results. Looking back at the error message, it's only really `pcs_new` that's causing the problem. The two main locations it's used is `input_writer.py` and `output_writer.py` at [`smac/util/io`](https://github.com/automl/SMAC3/tree/master/smac/utils/io).

I assume the errors is ConfigSpace doesn't know how to read or write categoricals with probabilities to a `pcs` file, which I assume is some format a config space is saved as.

To go about fixing it, I'm not sure but it would be good to look at an example pcs file and try to see if you can use `pcs_new` to write a categorical to file and then load it back in.

I may be wrong but that's my interpretation of the problem

hvarfner:
Alright! Everything else is now fixed (including some re-doing of tests for Normal parameters, where the defaults were off. See the latest commit and changes in tests for the best overview of the problem. I'll dig into it and see what I can do, but once that is resolved I have nothing more that I feel needs to be added.

hvarfner:
Also, does this mean that this feature is missing for the Beta parameters? It kind of seems like it, since I haven't touched that file.

mfeurer:
> I assume the errors is ConfigSpace doesn't know how to read or write categoricals with probabilities to a pcs file, which I assume is some format a config space is saved as.

Yes, indeed, these are files that define configuration spaces in a more human-readable way than JSON. They were defined prior to the existence of the ConfigSpace package. Both do not specify how to save probabilities for categoricals, so this is something which cannot be done. I believe it's time to disable saving PCS files in SMAC or at least not fail if it's not possible.

We should also make sure to better document the two converters and put links to the file definitions. Unfortunately, it's a few years since I last worked with these file types so I don't really remember where the definitions are. @mlindauer could you please help us out here?

hvarfner:
@mfeurer Interresting. Implementing the JSON part of reading/writing was not particularily difficult, as the pattern was very evident. This one, however, seems a lot more intricate. Does this feature need any treatment on the SMAC end of things, or is it enough to have it handled in ConfigSpace?

To be more clear: Can this be fixed by just adding stuff to ConfigSpace, and not SMAC?

ConfigSpace/configuration_space.pyx
```diff
@@ -1341,6 +1344,30 @@ class ConfigurationSpace(collections.abc.Mapping):
...
+        
+        Returns
+        -------
+        :class:`~ConfigSpace.configuration_space.ConfigurationSpace`
+            The resulting configuration space, without priors on the hyperparameters
+        """
+        uniform_types = (UniformFloatHyperparameter, UniformIntegerHyperparameter, OrdinalHyperparameter)
+        uniform_config_space = ConfigurationSpace()
+        for parameter in self.get_hyperparameters():
+            if isinstance(parameter, uniform_types):
```
mfeurer, line 1361: Could we check here whether the function `to_uniform` exists instead? Or add `to_uniform` to all hyperparameters, and they would just return a copy of themselves if they are uniform?

With the current code I would be worried that we forget to add new uniform types.

docs/source/User-Guide.rst
```diff
@@ -242,3 +242,45 @@ To read it from file
...
+>>> logmean = np.log(np.power(10.0, -3))
+>>> # two standard deviations on either side of the mean to cover the search space
+>>> logstd = np.log(10.0) 
+>>> learning_rate = CSH.NormalFloatHyperparameter(name='learning_rate', lower=1e-5, upper=1e-1, default_value=1e-3, mu=logmean, sigma=logstd, log=True)
+>>> dropout = CSH.BetaFloatHyperparameter(name='dropout', lower=0, upper=0.99, default_value=0.25, alpha=2, beta=4, log=False)
+>>> activation = CSH.CategoricalHyperparameter(name='activation', choices=['tanh', 'relu'], weights=[0.2, 0.8])
+
+>>> cs = ConfigurationSpace()
+>>> cs.add_hyperparameters([learning_rate, dropout, activation])
+Configuration space object:
```
mfeurer, line 268: Please check above how to activate doctests for these outputs.

ConfigSpace/hyperparameters.pyx
```diff
@@ -258,6 +258,38 @@ cdef class Constant(Hyperparameter):
                       transform: bool = False) -> List:
         return []
 
+    def pdf(self, vector: np.ndarray) -> np.ndarray:
+        """
+        Computes the probability density function of the parameter in 
+        the original parameter space (the one specified by the user).
+        For each parameter type, there is also a method _pdf which
```
mfeurer, line 265:
```suggestion
        For each hyperparameter type, there is also a method _pdf which 
```

ConfigSpace/hyperparameters.pyx
```diff
@@ -258,6 +258,38 @@ cdef class Constant(Hyperparameter):
...
 
+    def pdf(self, vector: np.ndarray) -> np.ndarray:
+        """
+        Computes the probability density function of the parameter in 
+        the original parameter space (the one specified by the user).
+        For each parameter type, there is also a method _pdf which 
+        operates on the transformed (and possibly normalized) parameter
+        space. Only legal values return a positive probability density,
+        otherwise zero.
+        vector: np.ndarray
```
mfeurer, line 269: Could you please adopt numpydoc docstring formatting?

ConfigSpace/hyperparameters.pyx
```diff
@@ -258,6 +258,38 @@ cdef class Constant(Hyperparameter):
...
+        type) space. As such, one never has to worry about log-normal 
+        distributions, only normal distributions (as the inverse_transform
+         in the pdf method handles these).
+        vector: np.ndarray
+            the (N, ) vector of imputs for which the probability density
+            function is to be computed.
+        """
+        return (vector == self.value).astype(float)
+
+    def get_max_density(self):
```
mfeurer, line 290: Hey, what's the usecase of this?

ConfigSpace/hyperparameters.pyx
```diff
@@ -258,6 +258,38 @@ cdef class Constant(Hyperparameter):
                       transform: bool = False) -> List:
         return []
 
+    def pdf(self, vector: np.ndarray) -> np.ndarray:
```
mfeurer, line 261: Could you please add `pdf` and `_pdf` to the `Hyperparameter` class, too?

mfeurer (commented):
Hi @hvarfner, I just briefly skimmed the PR and left a few first comments. I did not yet have the time to look through the PDF functions and the actual beta hyperparameter.

mfeurer:
FYI: I just turned on CI via github actions, so you can now see the test outcomes online.

ConfigSpace/hyperparameters.pyx
hvarfner: Good question! When running PiBO with RandomForests and continous parameters, one needs to discretize the pdf to avoid a pathological case where BO just gets stuck evaluating one point over and over. The short explanation for this is that a RF must have a piecewise continous acquisition function to actually work as intended.

To accomplish the discretized prior, it gets binned to certain threshold between its minimum and maximum. This is simply done by
 `np.linspace(0, max_value, some_number of bins)`
, where
`max_value = param.get_max_density()`
Hopefully, that clarifies it somewhat.

ConfigSpace/read_and_write/json.py
```diff
@@ -78,6 +80,17 @@ def _build_normal_float(param: NormalFloatHyperparameter) -> Dict:
     }
 
 
+def _build_beta_float(param: BetaFloatHyperparameter) -> Dict:
```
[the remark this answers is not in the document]
hvarfner: Sure!

ConfigSpace/hyperparameters.pyx
hvarfner: Sure, will fix!

mlindauer:
>  Unfortunately, it's a few years since I last worked with these file types so I don't really remember where the definitions are. @mlindauer could you please help us out here?

The newer format is defined here: https://bitbucket.org/mlindauer/aclib2/src/aclib2/AClib_Format.md
The old one should be part of the SMAC2 documentation.

hvarfner:
@mfeurer @eddiebergman If you don't mind, I'll take the liberty to create a workaround in SMAC to circumvent the issue of the PCS file. It is a bit beyond my current understanding to try and get it to work, and since Matthias said this was an option, I'll go ahead and propose something that addresses the issue.

hvarfner:
In the lastest PR for SMAC, there is a minor change to OutputWriter, where it only generates the PCS file if possible (otherwise, it just warns, like in the case of the categorical pdf). I believe this was the most harmless change possible, but it'd be good to know what you think on this. Otherwise, everything now runs as intended and the feedback from @mfeurer should be adressed - or at least I hope so.  I have a more sophisticated example in SMAC now, too, which you can try if you please.

Let me know if there's anything still missing, or if there's anything else you want to review! @eddiebergman @mfeurer

mfeurer (commented):
I made some further progress, I'm now at the beginning of the Beta hyperparameter.

ConfigSpace/configuration_space.pyx
```diff
@@ -1355,13 +1355,12 @@ class ConfigurationSpace(collections.abc.Mapping):
         :class:`~ConfigSpace.configuration_space.ConfigurationSpace`
             The resulting configuration space, without priors on the hyperparameters
         """
-        uniform_types = (UniformFloatHyperparameter, UniformIntegerHyperparameter, OrdinalHyperparameter)
         uniform_config_space = ConfigurationSpace()
         for parameter in self.get_hyperparameters():
-            if isinstance(parameter, uniform_types):
-                uniform_config_space.add_hyperparameter(parameter)
-            else:
+            if hasattr(parameter, 'to_uniform'):
```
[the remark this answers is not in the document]
hvarfner: Sure! I did not have this in mind, but it's probably best to do so, just to be sure.

ConfigSpace/hyperparameters.pyx
```diff
@@ -151,6 +151,36 @@ cdef class Hyperparameter(object):
...
+        For each parameter type, there is also a method _pdf which 
+        operates on the transformed (and possibly normalized) parameter
+        space. Only legal values return a positive probability density,
+        otherwise zero.
+        
+        Parameters
+        ----------
+        vector: np.ndarray
+            the (N, ) vector of imputs for which the probability density
+            function is to be computed.
```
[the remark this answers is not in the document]
hvarfner: Sure!

ConfigSpace/configuration_space.pyx
```diff
@@ -1355,13 +1355,12 @@ class ConfigurationSpace(collections.abc.Mapping):
...
                 uniform_config_space.add_hyperparameter(parameter.to_uniform())
+            else:
+                uniform_config_space.add_hyperparameter(parameter)
 
         uniform_config_space.add_conditions(self.get_conditions())
```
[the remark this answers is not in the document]
hvarfner: Hmm, okay. I don't understand the conditions perfectly, so can you please describe this a bit more in detail?

ConfigSpace/hyperparameters.pyx
```diff
@@ -151,6 +151,36 @@ cdef class Hyperparameter(object):
...
+        """
+        raise NotImplementedError()
```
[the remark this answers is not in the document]
hvarfner: @mfeurer I had to revisit this, since I thought about this quite a lot myself when deciding on the design.

Your suggestion is pretty much what I wanted to do. However, the _inverse_transform in constants and ordinals do not make sense with it, in my opinion. Therefore, I got a bit hesitant about the generality of the solution you're proposing. 
However, the workaround in this case is easy (or we just overwrite in that specific case, since the probability for ordinals is presumed uniform anyway, as of now). If you want to change it, I'm happy to do so!

The default would be to do like in FloatHP:
`        if vector.ndim != 1:`
            `raise ValueError("Method pdf expects a one-dimensional numpy array")`
        `vector = self._inverse_transform(vector)`
        `return self._pdf(vector)`

ConfigSpace/hyperparameters.pyx
```diff
@@ -151,6 +151,36 @@ cdef class Hyperparameter(object):
...
+        
+    def _pdf(self, vector: np.ndarray) -> np.ndarray:
+        """
+        Computes the probability density function of the parameter in 
+        the transformed (and possibly normalized, depends on the parameter 
+        type) space. As such, one never has to worry about log-normal 
+        distributions, only normal distributions (as the inverse_transform
+         in the pdf method handles these).
+        vector: np.ndarray
```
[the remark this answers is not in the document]
hvarfner: Fixing all of them

ConfigSpace/hyperparameters.pyx
```diff
@@ -258,6 +288,44 @@ cdef class Constant(Hyperparameter):
...
+        
+        Parameters
+        ----------
+        vector: np.ndarray
+            the (N, ) vector of imputs for which the probability density
+            function is to be computed.
+        """
+        if vector.ndim != 1:
+            raise ValueError("Method pdf expects a one-dimensional numpy array")
+        return self._pdf(vector)
```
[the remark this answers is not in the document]
hvarfner: @mfeurer 
Since inverse_transform returns a scalar in all cases, we can't. You'd have to redefine transform_scalar to return the same shape as the input (which seems reasonable to me), but I opted to do this instead. Happy to change it if you're OK with it.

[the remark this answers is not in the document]
hvarfner: So now I've worked around it instead. It's not as consistent as the alternative you propose (and why we can't have a unified pdf function for all HPs), but it yields correct results.

ConfigSpace/hyperparameters.pyx
```diff
@@ -374,6 +442,50 @@ cdef class FloatHyperparameter(NumericalHyperparameter):
...
+        space. Only legal values return a positive probability density,
+        otherwise zero.
+        
+        Parameters
+        ----------
+        vector: np.ndarray
+            the (N, ) vector of imputs for which the probability density
+            function is to be computed.
+        """
+        if vector.ndim != 1:
```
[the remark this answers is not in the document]
hvarfner: Yes, this is what the default would look like if I were to do the required changes to inverse_transform in Constant and Ordinal.

ConfigSpace/hyperparameters.pyx
```diff
@@ -374,6 +442,50 @@ cdef class FloatHyperparameter(NumericalHyperparameter):
...
+            raise ValueError("Method pdf expects a one-dimensional numpy array")
+        vector = self._inverse_transform(vector)
+        return self._pdf(vector)
+        
+    def _pdf(self, vector: np.ndarray) -> np.ndarray:
```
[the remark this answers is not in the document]
hvarfner: That was excessive. Duplicate, can remove.

ConfigSpace/hyperparameters.pyx
```diff
@@ -410,6 +522,52 @@ cdef class IntegerHyperparameter(NumericalHyperparameter):
...
+        ----------
+        vector: np.ndarray
+            the (N, ) vector of imputs for which the probability density
+            function is to be computed.
+        """
+        if vector.ndim != 1:
+            raise ValueError("Method pdf expects a one-dimensional numpy array")
+        is_integer = (np.round(vector) == vector).astype(int)
+        vector = self._inverse_transform(vector)
+        return self._pdf(vector) * is_integer
```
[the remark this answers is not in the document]
hvarfner: Yes. This is not paramount for any of the internals of PiBO to work, but it seemed like the correct thing to do. Do you want to change it?

(note that this would go away if we have a uniform pdf function. it can **not** be moved to _pdf.

ConfigSpace/hyperparameters.pyx
```diff
@@ -608,6 +766,20 @@ cdef class UniformFloatHyperparameter(FloatHyperparameter):
...
         return neighbors
 
+    def _pdf(self, vector: np.ndarray) -> np.ndarray:
+        # everything that comes into _pdf for a uniform variable should
+        # already be in [0, 1]-range, and if not, it's outside the upper
+        # or lower bound.
+        ub = 1
+        lb = 0
+        inside_range = ((lb <= vector) & (vector <= ub)).astype(int)
+        return inside_range / (self.upper - self.lower)
```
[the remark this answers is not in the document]
hvarfner: Fair point. Since not all inverse_transform work the same way (Uniform -> [0, 1], Normal -> [self._lower, self._upper], Beta -> [self._lower, self._upper], _pdf has to work with what it is given. Moreover, all FloatHPs have the same pdf, which forces the _pdf to get the user the expected output.

My approach has been to give the results that would be the most sensible to me if I were a user. As a user, I would expect a Uniform between 0 and 5 to have density 1/5 everywhere, and for that to be the case, _pdf has to look like this.

A similar discussion can be had about what the pdf of a logged uniform should output (arguably, it should be a log-uniform, but it's not). However, it is just uniform in our case, since I think that is what the user will have in mind.

ConfigSpace/hyperparameters.pyx
```diff
@@ -877,152 +1052,463 @@ cdef class NormalFloatHyperparameter(FloatHyperparameter):
             return np.rint((self.upper - self.lower) / self.q) + 1
 
 
-cdef class UniformIntegerHyperparameter(IntegerHyperparameter):
-    def __init__(self, name: str, lower: int, upper: int, default_value: Union[int, None] = None,
-                 q: Union[int, None] = None, log: bool = False,
+    def _pdf(self, vector: np.ndarray) -> np.ndarray:
+        if self.lower == None:
+            raise ValueError('Need upper and lower limits when using user priors.')
```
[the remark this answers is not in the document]
hvarfner: Regarding the quantization: It's transformed in the public method, so it is fine in the end, since we have sampling in the logged space, then the public method sample does the transform: possible exponentiation plus quantization.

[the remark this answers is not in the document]
hvarfner: For the latter: Yes, we can. I sort of defaulted on that one.

ConfigSpace/hyperparameters.pyx
```diff
@@ -877,152 +1052,463 @@ cdef class NormalFloatHyperparameter(FloatHyperparameter):
...
         -------
 
-        >>> import ConfigSpace as CS
-        >>> import ConfigSpace.hyperparameters as CSH
-        >>> cs = CS.ConfigurationSpace(seed=1)
-        >>> uniform_integer_hp = CSH.UniformIntegerHyperparameter(name='uni_int', lower=10,
-        ...                                                       upper=100, log=False)
-        >>> cs.add_hyperparameter(uniform_integer_hp)
-        uni_int, Type: UniformInteger, Range: [10, 100], Default: 55
+            >>> import ConfigSpace as CS
```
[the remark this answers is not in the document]
hvarfner: Will check & change.

ConfigSpace/hyperparameters.pyx
```diff
@@ -1138,7 +1632,8 @@ cdef class UniformIntegerHyperparameter(IntegerHyperparameter):
 cdef class NormalIntegerHyperparameter(IntegerHyperparameter):
     cdef public mu
     cdef public sigma
-    cdef nfhp
+    cdef public nfhp
+    cdef normalization_constant
```
mfeurer, line 1636: Is there a reason we don't have this for the normal float hyperparameter?

ConfigSpace/hyperparameters.pyx
```diff
@@ -1521,6 +2350,16 @@ cdef class CategoricalHyperparameter(Hyperparameter):
             meta=self.meta
         )
 
+    # This is just for the uniform configspace for PiBO
```
mfeurer, line 2353: Could you please add a link to PiBO?

ConfigSpace/hyperparameters.pyx
```diff
@@ -1605,10 +2444,13 @@ cdef class CategoricalHyperparameter(Hyperparameter):
         except ValueError:
             return None
 
-    def _inverse_transform(self, vector: Union[None, str, float, int]) -> Union[int, float]:
+
+    # TEncountered something of a bug here when this was passed through .pdf
```
mfeurer, line 2448:
```suggestion
    # Encountered something of a bug here when this was passed through .pdf
```

ConfigSpace/hyperparameters.pyx
```diff
@@ -1605,10 +2444,13 @@ cdef class CategoricalHyperparameter(Hyperparameter):
...
+    # for some reason, it wanted a numpy array as output (and as such, it is now changed)
+    def _inverse_transform(self, vector: Union[None, np.ndarray, str, float, int]) -> Union[np.ndarray, int, float]:
```
mfeurer, line 2450:
```suggestion
    def _inverse_transform(self, vector: Union[None, np.ndarray, str, float, int]) -> Union[np.ndarray, float]:
```

mfeurer (commented):
And some more comments. I still didn't look through the BetaIntegerHyperparameter and test_hyperparameters.py. Nevertheless, I think this PR now contains at least 3 atomic features/changes:
1. A few bug fixes
2. get_pdf and get_max_density
3. new distributions
Because this PR is getting prettly large and these three changes intertwine quite a bit I'm wondering whether we should split this up into three separate PRs. @eddiebergman what's your thought on this? @hvarfner would you be okay creating three shorter PRs, each focusing on one aspect/issue?

ConfigSpace/hyperparameters.pyx
hvarfner: The normalization constant? The continous distribution is already normalized per default, whereas the discrete one is not. 

If we wanted the build-in normalization in the discrete case, we'd have to use a binomial. However, since a continuous normal was used before to power the normal integer HP, I decided to stick with what was already there.

ConfigSpace/hyperparameters.pyx
hvarfner: Sure

ConfigSpace/hyperparameters.pyx
```diff
@@ -877,152 +1052,463 @@ cdef class NormalFloatHyperparameter(FloatHyperparameter):
...
+            q=self.q,
+            lower=self.lower,
+            upper=self.upper,
+            meta=self.meta
+        )
 
-    def is_legal(self, value: int) -> bool:
-        if not (isinstance(value, (int, np.int32, np.int64))):
+    def __hash__(self):
+        return hash((self.name, self.alpha, self.beta, self.log, self.q))
```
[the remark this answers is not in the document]
hvarfner: No. I'll add it.

ConfigSpace/hyperparameters.pyx
```diff
@@ -877,152 +1052,463 @@ cdef class NormalFloatHyperparameter(FloatHyperparameter):
...
+                                          ub,
+                                          default_value=self.default_value,
+                                          q=self.q, log=self.log)
+
+    def check_default(self, default_value: Union[int, float]) -> Union[int, float]:
+        # return mode as default
+        lb = self.lower
+        ub = self.upper
+        
+        if (default_value is None) or (default_value < lb) or (default_value > ub):
```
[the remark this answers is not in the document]
hvarfner: No, it was simply to be nice to the user if they entered something that was out of bounds. I will change it so that it raises an error instead.

ConfigSpace/hyperparameters.pyx
```diff
@@ -877,152 +1052,463 @@ cdef class NormalFloatHyperparameter(FloatHyperparameter):
...
+        elif self.is_legal(default_value):
+            return default_value
+        else:
+            raise ValueError("Illegal default value %s" % str(default_value))
+
+    def to_integer(self) -> 'BetaIntegerHyperparameter':
+        if self.q is None:
+            q_int = None
+        else:
+            q_int = int(self.q)
```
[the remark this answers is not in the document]
hvarfner: yes

PR #219, opened 2022-01-21
Diamond cond with OrConjunction 

dengdifan:
This PR works on solving a potential error that might occur for a hierarchy configuration space with OrConjunction:
Suppose that an HP `top` has two choices `A` and `B` and an HP `bottom` that is conditioned on `A` or `B`: `(bottom | top == A || top == B)`. Then if `top` switches from `A` to `B`,  `bottom` should remain to activate (A more concrete example could be found under test/test_util.py::UtilTest::test_check_neighbouring_config_diamond_or_conjunction).

An additional check is attached under `change_hp_value`. However, this might make the check slower: we need to check all the children conditions and not skip them as before

### DIFF
--- a/ConfigSpace/c_util.pyx
+++ b/ConfigSpace/c_util.pyx
@@ -288,7 +288,8 @@ cpdef np.ndarray change_hp_value(
     # Hyperparameters which are going to be set to inactive
     disabled = []
 
-    # Hyperparameters which are going to be set activate, this is applicable to diamond conditional space:
+    # Hyperparameters which are going to be set activate, we introduce this to resolve the conflict that might be raised
+    # by OrConjunction:
     # Suppose that we have a parent HP_p whose possible values are A, B, C; a child HP_d is activate if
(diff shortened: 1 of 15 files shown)


ConfigSpace/c_util.pyx
```diff
@@ -315,6 +322,16 @@ cpdef np.ndarray change_hp_value(
                     active = False
                     break
 
+            if active:
+                hps_to_be_activate.add(current_idx)
+                if current_value == current_value:
```
eddiebergman, line 327: This seems odd? `current_value == current_value`, is this not always true?

eddiebergman (commented):
Just need to ask about the odd `==` thing. Otherwise a rough understanding of the code and the test seem to be correct.

ConfigSpace/c_util.pyx
dengdifan: Not really, if current_value is deactivated, then this equation does not hold (Similar code can be found under line 335 and 344)

PR #221, opened 2022-01-25
PiBO Pull Request 1 - Bugfixes

hvarfner:
# PiBO Pull Request 1 - Bugfixes

#### Bugfix - Normal HPs
- Fixed sampling in Normal HPs - did not sample to the correct distribution for logged variables previously
- Fixed the definition of parameters in logged Normal HPs - now, they follow a log-normal distribution, so mu=1 for a logged variable gived mode at e, mu=0 gives mode at 1
- Fixed all tests accordingly

#### Bugfix - to_uniform
- Removed unnecesary rounding for float HPs

#### Bugfix - to_integer
- Changed incorrect int() rounding, which produces same result as np.floor().astype(int)
- Now, the lower value of the float is rounded up (so lower=-1.5 --> -1 for integerHP, and upper is rounded down: 6.5 --> 6 (which is obiously guaranteed to be legit, whereas rounding down in both cases does not work.

#### Change - Categorical HPs
- Categorical HPs always have probabilities (uniform if nothing else is specified)
- Changed tests accordingly


ConfigSpace/hyperparameters.pyx
```diff
@@ -1483,7 +1493,8 @@ cdef class CategoricalHyperparameter(Hyperparameter):
         repr_str.write("}")
         repr_str.write(", Default: ")
         repr_str.write(str(self.default_value))
-        if self.probabilities is not None:
+        # if the probability distribution is not uniform, write out the probabilities
+        if not np.all(self.probabilities == self.probabilities[0]):
```
eddiebergman, line 1497: You removed the condition `if self.probabilities is not None:`, is `self.probabilities` guaranteed to exist now?

ConfigSpace/hyperparameters.pyx
```diff
@@ -1521,6 +1532,16 @@ cdef class CategoricalHyperparameter(Hyperparameter):
             meta=self.meta
         )
 
+    # This is just used for the uniform configspace for PiBO: https://openreview.net/forum?id=MMAeCXIa89
+    # It creates a categorical parameter with equal weights for all choices
```
eddiebergman, line 1536: Add it as a docstring and not a comment

test/test_hyperparameters.py
```diff
@@ -464,29 +463,27 @@ def test_normalint(self):
             "param, Type: NormalInteger, Mu: 0 Sigma: 10, Default: 0, Q: 2",
             str(f2_large_q))
 
-        with pytest.warns(RuntimeWarning, match='divide by zero encountered in log'):
-            f3 = NormalIntegerHyperparameter("param", 0, 10, log=True)
-        with pytest.warns(RuntimeWarning, match='divide by zero encountered in log'):
-            f3_ = NormalIntegerHyperparameter("param", 0, 10, log=True)
+        f3 = NormalIntegerHyperparameter("param", 0, 10, log=True)
+        f3_ = NormalIntegerHyperparameter("param", 0, 10, log=True)
```
eddiebergman, line 467: I'm okay with removing the warning check as it's not something we explicitly do, i.e. emit a warning. Just wanted to ask if the warning is still raised however or was this handled in some other way such that no warning happens?

eddiebergman (commented):
Seems fine, minor comments but happy with it :)

ConfigSpace/hyperparameters.pyx
hvarfner: Yes!

mfeurer:
Hey @hvarfner I'll have a look once the tests are green, thanks @eddiebergman for pointing Carl to the test setup. 

WRT the PCS format: I'd be fine with merging this PR into master, if we do a squash-merge, rolling back would be easy in case things break.

hvarfner:
Alright, thanks! I will get on the tests, and fix what you proposed!

test/test_hyperparameters.py
```diff
@@ -926,7 +923,7 @@ def actual_test():
                 counts_per_bin[value] += 1
 
             self.assertEqual(
-                {0: 2456, 2: 2485, 'Bla': 2550, u'Blub': 2509},
+                {0: 2539, 2: 2451, 'Bla': 2549, 'Blub': 2461},
```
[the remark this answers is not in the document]
hvarfner: As for this, the change did indeed occur when I added the default weights (or, I'm 90% sure that's the case. I'll check).

test/test_hyperparameters.py
hvarfner: This has to do with the fact that I changed what the parameters mu and sigma actually mean for a logged normal variable.

Previously, the mean value would be defined as it it was a Normal, and not a LogNormal. The default value would be defined as log(mu), i.e. log(0) --> the warning occurs. This was simply a bug, as it was not correctly defined.

Now, the parameters work as they do in a LogNormal distribution. As such, the user enterfing mu = 1 -->  mean of distribution at e. While this is a bit unintuitive, I would argue that any other solution is to some extent incorrect and leaves the user guessing. 

As for the warning, it simply never occurs now when the definition is changed.

ConfigSpace/hyperparameters.pyx
hvarfner: Sure

ConfigSpace/hyperparameters.pyx
```diff
@@ -1521,6 +1532,16 @@ cdef class CategoricalHyperparameter(Hyperparameter):
             meta=self.meta
         )
 
+    # This is just used for the uniform configspace for PiBO: https://openreview.net/forum?id=MMAeCXIa89
```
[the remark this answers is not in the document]
hvarfner: Ok

test/test_hyperparameters.py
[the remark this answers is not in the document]
hvarfner: Confirmed this is the case.

eddiebergman (approved):

mfeurer (changes requested):
The changes themselves look good to me. However, the json serialization does not work, and the test for that fails. I assume that this is because of the new weights of the categorical hyperparameter, but I'm not sure. Could you please have a look into this?

hvarfner:
@mfeurer Sorry, but do some tests fail for you? All tests pass for me, so it would be great if you could provide some additional info.

mfeurer:
Good morning @hvarfner, I'm referring to the tests run on github actions. You can see them between your last answer and the answer button:
image
Right now it says 38 failing and 1 successful checks. If you click on one of the 'Details' links you will find further information on what's going wrong. I browsed them yesterday, and it appears that only a single unit test is failing (but across all 38 different setups we test for).

hvarfner:
@mfeurer  Right, got it! Thanks. I'll dig into them - I've obviously been ignoring them so far. Is there any way to test them locally?

eddiebergman:
You can generally install the repo locally and run the tests
```python
git clone git@github.com:hvarfner/ConfigSpace.git
cd ConfigSpace
# ... make virtaul env
pip install -e ".[tests]"
pytest
```

* Check out `pytest --help` for more information on how to select certain tests.
* A useful reference we have in [`CONTRIBUTINg.md` for autosklearn](https://github.com/automl/auto-sklearn/blob/master/CONTRIBUTING.md#testing)
* The `-e` argument above means it install the library in "editable" mode, meaning you can edit the code and run the tests, with your edited changes taking place.

hvarfner:
Hi @eddiebergman ,

I tried it, and I think (I'm really no expert at this whatsoever) that was indeed what I was running all along. It's a bit difficult to try and fix things without really understanding what it can be. Since I changed some of the tests (due to no support for PCS & categorical), does that have something to do with it? 
image

eddiebergman:
Have you pushed your latest changes to github, doesn't seem to have been updated in a while.

Running that specific check that is failing is
```
python -k "test_round_trip"
```

It seems it's to do with config equality, there was a [change 20 days ago](https://github.com/automl/ConfigSpace/commit/f1ead4c329abf5004970530882ac782766173474) that may impact it, the main one being to do with equaility now considered the `default_value`. You could try merging with master and seeing if that changes things.

hvarfner:
Thanks for the quick reply. The latest commit was 4 days ago, so everything is up to date.
So, merging with master did not solve the issue (or well, since master has not yet accounted for the fails in PCS, some more tests failed, but test_round_trip was not one of them).

If test_round_trip is the issue, that would make sense with the changes to the PCS for categoricals. However, I still fail to see why it passes for me locally but not here.

eddiebergman:
Can you go to a fresh location, fresh virtual env and do a fresh install of this repo/branch from github, i.e. don't use your local copy at all. If it doesn't fail there then I have no idea.

hvarfner:
Unfortunately, no improvement. I'm unable to get the test_round_trip test to fail, and so I really don't know how to approach this. The assertion is made on a pair of 32 000 token strings (both of which have the same length in the failed test), so I really don't know on this one. Do you have any idea? Is the source code different for the tests?

eddiebergman:
I ran this branch locally and got no test failures either. I will rerun these github action tests, if they fail again, I'm willing to push this through anyways and just log that this happened as an Issue.

hvarfner:
Thanks, much appreciated!

eddiebergman:
I did some digging and it seems our automated tests are pulling your latest commit on this PR, they're pulling [this one](https://github.com/automl/ConfigSpace/commit/5fdf7ac19b5e95910996676b428adb257cc17fde). I have no idea why that is ... However I can confirm your latest commit is actually correct so I would vote to disregard these checks but try and figure out why it pulls the merge commit instead of the most recent.

@mfeurer thoughts?

eddiebergman:
Can you try one last thing, just push an empty commit
```
git commit --allow-empty -m "Fix?"
```

hvarfner:
Sorry, was away for lunch. It's done!

eddiebergman:
No problem, thanks for that! It appears it's still pulling the old version and isn't pulling the latest commit. I'm going to merge this and finally progress on to stage 2. If the error crops up there again then we will need to somehow fix it but hopefully it dissapears.

eddiebergman:
The failed pcs file is `SparrowToRiss-cssc14.pcs` which is read in, written out and then read back in and then check for equality, somewhere this is failing, I'm assuming this is to do with the pcs changes that are in the next PR?

eddiebergman:
I check through the other 2 PR's, it doesn't seem the pcs file format was fixed at any point?

hvarfner:
Well, the PCS format was not fixed, but rather disabled for ConfigSpaces that include categoricals.

eddiebergman:
I'm not sure that's something we can really do? It basically means ConfigSpace won't allow any kind of exporting for people that use categoricals. I think we would need some solution to this by the end of this PR.

@mfeurer

hvarfner:
Well, since every categorical has weights with this solution, we'd have to make categoricals only have weights in the case where the user specifies it. It seemed like a less consistent solution in my opinion, but I guess it can be changed if you deem it to be the best solution.

mfeurer:
Good morning. I'm on this. The bug appears because for the json serialization of the ConfigSpace `SparrowToRiss-cssc14.pcs` we loose some precision in the hyperparameter `init-act`:

```
From PCS: (0.14285714285714285, 0.14285714285714285, 0.14285714285714285, 0.14285714285714285, 0.14285714285714285, 0.14285714285714285, 0.14285714285714285)
From JSON: (0.14285714285714288, 0.14285714285714288, 0.14285714285714288, 0.14285714285714288, 0.14285714285714288, 0.14285714285714288, 0.14285714285714288)
```
Now the question is: what's going on here? I'll keep you posted.

mfeurer:
And the issue is https://github.com/automl/ConfigSpace/blob/f1ead4c329abf5004970530882ac782766173474/ConfigSpace/hyperparameters.pyx#L1598

The hyperparameter is serialized to json, then de-serialized from json. Until then everything is fine. Now when the constructor is called it invokes the function `_get_probabilities`, which in turn tries to normalize the probabilities, creating this numeric instability. The actual issue might be the fact that we renormalize the probabilities, or that the probabilities don't sum up to 1 in the first place.

Potentially we need to store and serialize the weights, and have the probabilities as an internal representation only. Then the probabilities will only be computed on `__init__` from the original weights, and should always be equal. What do you think?

hvarfner:
If you're asking me, I think that seems like a no-brainer. I don't really see why the probabilities need to be stored externally if the weights are already there.

mfeurer:
> If you're asking me, I think that seems like a no-brainer. I don't really see why the probabilities need to be stored externally if the weights are already there.

Great :) Could you please go ahead and store and serialize the weights?

hvarfner:
Sure thing!

mfeurer:
To simplify testing you can temporarily prune the file `SparrowToRiss-cssc14.pcs ` to only contain:
```
init-act  {0,1,2,3,4,5,6}[1]       # initialize activities (0=none,1=inc-lin,2=inc-geo,3=dec-lin,4=dec-geo,5=rnd,6=abs(jw))
init-pol  {0,1,2,3,4,5}[2]       # initialize polarity (0=none,1=JW-pol,2=JW-neg,3=MOMS,4=MOMS-neg,5=rnd)
```
This will continue to fail as long as the bug is there.

hvarfner:
Okay, I guess we're getting closer. With the remaining errors, and PCS files, I would like some input as to what is actually the issue, and what to do with the categorical HPs in the PCS format. Opinions?

ConfigSpace/hyperparameters.pyx
```diff
@@ -1476,6 +1477,7 @@ cdef class CategoricalHyperparameter(Hyperparameter):
             raise TypeError('Using a set of weights is prohibited as it can result in '
                             'non-deterministic behavior. Please use a list or a tuple.')
         self.choices = tuple(choices)
+        self.weights = weights
```
mfeurer, line 1480:
```suggestion
        self.weights = tuple(weights)
```

ConfigSpace/hyperparameters.pyx
```diff
@@ -1407,11 +1407,12 @@ cdef class NormalIntegerHyperparameter(IntegerHyperparameter):
 
 cdef class CategoricalHyperparameter(Hyperparameter):
     cdef public tuple choices
-    cdef public tuple probabilities
+    cdef public list weights
```
mfeurer, line 1410:
```suggestion
    cdef public tuple weights
```

mfeurer (commented):
The tests look good and we can take care of the failing doctest after this PR is merged. As a last step I would suggest replacing the `list` of weights by `tuple` of weights because they are immutable.

ConfigSpace/hyperparameters.pyx
hvarfner: @mfeurer Had to make one more fix in order to make this work, but now it does!

mfeurer (approved):

PR #223, opened 2022-01-25
PiBO Pull Request 3 - PDFs and Densities

hvarfner:
# PiBO Pull Request 3 - PDFs and Densities

### Third pull request - Implements _pdf, pdf and get_max_density for each parameter type, as well as other support needed for PiBO

- Everything from PiBO PR1 and PR2
- _pdf and pdf
- get_max_density
- remove_parameter_priors with supporting static methods which copy forbiddens and conditions from the old (non-uniform) configspace to the new one
    - static method substitute_hyperparameters_in_conditions() 
    - static method substitute_hyperparameters_in_forbiddens()
- Accompanying tests
- Documentation in user guide


eddiebergman:
Hi @hvarfner, I can review this now before review 2 is corrected if you like, else we can wait for PR 2 to be fixed, let me know :)

hvarfner:
@eddiebergman If you could, that would be great. These two are fairly disconnected, so I think that's a good idea!

PR #230, opened 2022-01-31
Allow using forbidden clauses with non-numeric values

filipbartek:
# Before

Restricting a categorical hyperparameter with non-numeric (namely string) values by a forbidden clause `forbidden` and calling `forbidden.is_forbidden` on a relevant `instantiated_hyperparameters` dictionary raises the following exception:

```
TypeError: must be real number, not str
```

# After

Forbidden clauses that involve non-numeric hyperparameters can be evaluated using `is_forbidden` without raising an exception.

### DIFF
--- a/ConfigSpace/forbidden.pyx
+++ b/ConfigSpace/forbidden.pyx
@@ -154,7 +154,7 @@ cdef class SingleValueForbiddenClause(AbstractForbiddenClause):
 
         return self._is_forbidden_vector(value)
 
-    cdef int _is_forbidden(self, float value):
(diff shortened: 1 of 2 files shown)


mfeurer:
Thanks for catching this. I can reproduce the issue and am currently checking your proposed fix. It looks good. 

I have one question, though: it appears that `Condition.is_forbidden` is only invoked by unit tests. How did you find this issue?

PS: Your other PR is next on my list.

mfeurer:
Got it, thanks for the explanation. I could find no performance decrease when benchmarking the sampling speed, so let's merge this :)

Issue #198, opened 2021-09-22
Constraint between 2 hyperparameters


mhoangvslev:
Hi,

Given the config below:

```python
{
    "batch_size": 32,
    "clipping_value": 0.011738590592399789,
    "discount": 0.9487457605181253,
    "entropy_regularization": 0.5422291367235231,
    "exploration": "exponential",
    "exploration[exponential][decay]": "polynomial",
    "exploration[exponential][decay_rate]": 0.08594089681763346,
    "exploration[exponential][initial_value]": 0.40700367119366254,
    "exploration[exponential][num_steps]": 10332,
    "exploration[exponential][unit]": "episodes",
    "learning_rate": 0.050051756778602824,
    "memory": 10000,
    "update_frequency": 64
}
```

`update_frequency` must always be lesser than `batch_size`, therefore the above config is discarded after the referenced code raised an error. Is there a way to constraint the values of 2 hyperparams?

mfeurer:
We currently do not support. This feature would require adding a new kind of forbidden clause to https://github.com/automl/ConfigSpace/blob/master/ConfigSpace/forbidden.pyx

Ennosigaeon:
@mhoangvslev I stumbled across the same limitation and decided to give the implementation a try. Would #245 solve your problem?

mfeurer:
Will be available in the next release.

PR #245, opened 2022-04-01
Add: Add forbidden relations between two hyperparameters

Ennosigaeon:
Add forbidden relations between two hyperparameters. It is possible to define less than, equals and greater than relations.

This PR solves #198.

I am not sure about necessary adaption for `read_write.pcs`, I would be great if you could give me a pointer how to implement that correctly. If I forgot some other place, please just let me know.

Short disclaimer: I haven't written python extensions for a long time, please be careful checking that I didn't mess up...

### DIFF
--- a/ConfigSpace/forbidden.pyx
+++ b/ConfigSpace/forbidden.pyx
@@ -133,7 +133,7 @@ cdef class SingleValueForbiddenClause(AbstractForbiddenClause):
         if value is None:
             if strict:
                 raise ValueError("Is_forbidden must be called with the "
-                                 "instanstatiated hyperparameter in the "
+                                 "instantiated hyperparameter in the "
                                  "forbidden clause; you are missing "
                                  "'%s'" % self.hyperparameter.name)
             else:
@@ -146,7 +146,7 @@ cdef class SingleValueForbiddenClause(AbstractForbiddenClause):
         if value != value:
             if strict:
                 raise ValueError("Is_forbidden must be called with the "
-                                 "instanstatiated vector id in the "
+                                 "instantiated vector id in the "
                                  "forbidden clause; you are missing "
                                  "'%s'" % self.vector_id)
             else:
@@ -188,7 +188,7 @@ cdef class MultipleValueForbiddenClause(AbstractForbiddenClause):
         if value is None:
             if strict:
                 raise ValueError("Is_forbidden must be called with the "
-                                 "instanstatiated hyperparameter in the "
+                                 "instantiated hyperparameter in the "
                                  "forbidden clause; you are missing "
                                  "'%s'." % self.hyperparameter.name)
             else:
@@ -202,7 +202,7 @@ cdef class MultipleValueForbiddenClause(AbstractForbiddenClause):
         if value != value:
             if strict:
                 raise ValueError("Is_forbidden must be called with the "
-                                 "instanstatiated vector id in the "
+                                 "instantiated vector id in the "
                                  "forbidden clause; you are missing "
                                  "'%s'" % self.vector_id)
             else:
@@ -378,7 +378,7 @@ cdef class AbstractForbiddenConjunction(AbstractForbiddenComponent):
             if dlc.hyperparameter.name not in ihp_names:
                 if strict:
                     raise ValueError("Is_forbidden must be called with all "
-                                     "instanstatiated hyperparameters in the "
+                                     "instantiated hyperparameters in the "
                                      "and conjunction of forbidden clauses; "
                                      "you are (at least) missing "
                                      "'%s'" % dlc.hyperparameter.name)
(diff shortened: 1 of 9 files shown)


test/test_forbidden.py
```diff
@@ -225,3 +225,36 @@ def test_and_conjunction(self):
             self.assertEqual(results[i], is_forbidden)
 
             self.assertFalse(total_and.is_forbidden({}, strict=False))
+
+    def test_relation(self):
+        hp1 = CategoricalHyperparameter("cat_int", [0, 1])
+        hp2 = UniformIntegerHyperparameter("int", 0, 2)
+        hp3 = UniformIntegerHyperparameter("int2", 0, 2)
+        hp4 = UniformFloatHyperparameter("float", 0, 2)
+        hp5 = CategoricalHyperparameter("str", ['foo', 'bar'])
```
mfeurer, line 234: Seeing this, I'm wondering whether we should forbid categorical and ordinal hyperparameters for now or add documentation that the forbiddens can result in weird behavior here? Or we'd need to think how to implement less than and greater than correctly for ordinals and categorical (as they can behave differently when they are compared with their transformed version (_is_forbidden_vector) than their untransformed version (_is_forbidden)).

ConfigSpace/forbidden.pyx
```diff
@@ -493,3 +493,202 @@ cdef class ForbiddenAndConjunction(AbstractForbiddenConjunction):
...
+
+        self.left = left
+        self.right = right
+        self.vector_ids = (-1, -1)
+
+    def __eq__(self, other: Any) -> bool:
+        if not isinstance(other, self.__class__):
+            return False
+
+        return (self.left.name == other.left.name and
```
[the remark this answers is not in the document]
Ennosigaeon: I can't recall a particular reason. Removed `.name`

ConfigSpace/forbidden.pyx
```diff
@@ -493,3 +493,202 @@ cdef class ForbiddenAndConjunction(AbstractForbiddenConjunction):
...
+        if right != right:
+            if strict:
+                raise ValueError("Is_forbidden must be called with the "
+                                 "instantiated vector id in the "
+                                 "forbidden clause; you are missing "
+                                 "'%s'" % self.vector_ids[1])
+            else:
+                return False
+
+        return self._is_forbidden(self.left._transform(left), self.right._transform(right))
```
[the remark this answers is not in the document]
Ennosigaeon: Changed

ConfigSpace/forbidden.pyx
```diff
@@ -493,3 +493,202 @@ cdef class ForbiddenAndConjunction(AbstractForbiddenConjunction):
...
+     left : :ref:`Hyperparameters`
+         left side of the comparison
+     right : :ref:`Hyperparameters`
+         right side of the comparison
+    """
+
+    def __repr__(self):
+        return "Forbidden: %s < %s" % (self.left.name, self.right.name)
+
+    cdef _is_forbidden(self, left, right):
```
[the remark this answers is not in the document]
Ennosigaeon: I added the return type to all functions below

test/test_forbidden.py
Ennosigaeon: I specifically require the option to compare a list of integers with an int hyperparameter. Therefore, my idea was to check the compatibility when evaluating the actual configuration. In case two hyperparameters are not comparable python raises a TypeError. That's the reason why I check for a TypeError in the test case below when comparing `1 == 'foo'`.

Regarding the problem with `_is_forbidden_vector`: I explicitly convert from the vector representation to the represented value in line #580 to circumvent this problem. So basically `_is_forbidden_vector` behaves like `_is_forbidden`.

If you require to validate the configuration space during construction, I can adapt the code to only accept NumericalHyperparamter.

test/test_forbidden.py
```diff
@@ -225,3 +225,36 @@ def test_and_conjunction(self):
...
+        self.assertFalse(forb.is_forbidden({'int': 1, 'int2': 1}, True))
+        self.assertFalse(forb.is_forbidden({'int': 0, 'int2': 1}, True))
+
+        forb = ForbiddenGreaterThan(hp3, hp4)
+        self.assertTrue(forb.is_forbidden({'int2': 1, 'float': 0}, True))
+        self.assertFalse(forb.is_forbidden({'int2': 1, 'float': 1}, True))
+        self.assertFalse(forb.is_forbidden({'int2': 0, 'float': 1}, True))
+
+        forb = ForbiddenGreaterThan(hp4, hp5)
+        self.assertRaises(TypeError, forb.is_forbidden, {'float': 1, 'str': 'foo'}, True)
```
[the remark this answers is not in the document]
Ennosigaeon: See comment above

Ennosigaeon:
@mfeurer I added a check in `pcs_new.write` that prevents serializing ConfigSpaces containing ForbiddenRelations. `pcs` should be save as it only supports `ForbiddenEqualsClause` and `ForbiddenAndConjunction` (if I understand the code correctly).

All changes are in a separate commit that I will squash once you are happy.

ConfigSpace/forbidden.pyx
```diff
@@ -493,3 +493,202 @@ cdef class ForbiddenAndConjunction(AbstractForbiddenConjunction):
...
+        return left < right
+
+    cdef int _is_forbidden_vector(self, DTYPE_t left, DTYPE_t right):
+        return left < right
+
+
+cdef class ForbiddenEquals(ForbiddenRelation):
```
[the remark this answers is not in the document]
Ennosigaeon: I changed all names accordingly

test/test_forbidden.py
mfeurer: Okay, got it. So the behavior basically is:
* int and float: all fine
* categorical or ordinal that contains int or float: all fine
* categorical or ordinal that contain strings: ordering is determined based on a string comparison, not on how the hyperparameter type should treat it (categorical should always return false for larger or smaller, ordinal should consider the ordinal relation).
If that's the case it would be great if you could add this to the docstring so that users aren't surprised.

ConfigSpace/forbidden.pyx
```diff
@@ -600,6 +601,12 @@ cdef class ForbiddenLessThanRelation(ForbiddenRelation):
     >>> cs.add_forbidden_clause(forbidden_clause)
     Forbidden: a < b
 
+    Note
+    ____
```
mfeurer, line 605:
```suggestion
    ----
```

ConfigSpace/forbidden.pyx
```diff
@@ -637,6 +644,12 @@ cdef class ForbiddenEqualsRelation(ForbiddenRelation):
     >>> cs.add_forbidden_clause(forbidden_clause)
     Forbidden: a == b
 
+    Note
+    ____
```
mfeurer, line 648:
```suggestion
    ----
```

ConfigSpace/forbidden.pyx
```diff
@@ -674,6 +687,12 @@ cdef class ForbiddenGreaterThanRelation(ForbiddenRelation):
     >>> cs.add_forbidden_clause(forbidden_clause)
     Forbidden: a > b
 
+    Note
+    ____
```
mfeurer, line 691:
```suggestion
    ----
```

ConfigSpace/forbidden.pyx
```diff
@@ -600,6 +601,12 @@ cdef class ForbiddenLessThanRelation(ForbiddenRelation):
     >>> cs.add_forbidden_clause(forbidden_clause)
     Forbidden: a < b
 
+    Note
+    ____
+    If the values of the both hyperparameters are not comparible
+    (e.g. comapring int and str), a TypeError is raised. For OrdinalHyperparameters
```
mfeurer, line 607:
```suggestion
    (e.g. comparing int and str), a TypeError is raised. For OrdinalHyperparameters
```

ConfigSpace/forbidden.pyx
```diff
@@ -600,6 +601,12 @@ cdef class ForbiddenLessThanRelation(ForbiddenRelation):
...
+    the actual values are used for comparison *not* their oridnal value.
```
mfeurer, line 608:
```suggestion
    the actual values are used for comparison **not** their ordinal value.
```

ConfigSpace/forbidden.pyx
```diff
@@ -637,6 +644,12 @@ cdef class ForbiddenEqualsRelation(ForbiddenRelation):
     >>> cs.add_forbidden_clause(forbidden_clause)
     Forbidden: a == b
 
+    Note
+    ____
+    If the values of the both hyperparameters are not comparible
+    (e.g. comapring int and str), a TypeError is raised. For OrdinalHyperparameters
```
mfeurer, line 650:
```suggestion
    (e.g. comparing int and str), a TypeError is raised. For OrdinalHyperparameters
```

ConfigSpace/forbidden.pyx
```diff
@@ -637,6 +644,12 @@ cdef class ForbiddenEqualsRelation(ForbiddenRelation):
...
+    the actual values are used for comparison *not* their oridnal value.
```
mfeurer, line 651:
```suggestion
    the actual values are used for comparison **not** their ordinal value.
```

ConfigSpace/forbidden.pyx
```diff
@@ -674,6 +687,12 @@ cdef class ForbiddenGreaterThanRelation(ForbiddenRelation):
     >>> cs.add_forbidden_clause(forbidden_clause)
     Forbidden: a > b
 
+    Note
+    ____
+    If the values of the both hyperparameters are not comparible
+    (e.g. comapring int and str), a TypeError is raised. For OrdinalHyperparameters
+    the actual values are used for comparison *not* their oridnal value.
```
mfeurer, line 694:
```suggestion
    the actual values are used for comparison **not** their ordinal value.
```

test/test_forbidden.py
mfeurer: Thanks a lot for the explanation and the updated tests, that makes sense. Would you mind adding a case for Ordinal that covers a hyperparameter like `OrdinalHyperparameter("water_temperature", ["cold", "luke-warm", "hot", "boiling"])` to test that we're failing or comparing according to string comparisons, not the actual comparisons?

Ennosigaeon:
I have added two extra asserts in the relation test. Yet, I am not completely sure, whether this is the test you intended.

mfeurer (approved):
Yes, that test is exactly as I expected it to look like.

I just fixed two PEP8 issues as I'm otherwise very happy with this PR and will merge now.

Ennosigaeon:
Do you want me to squash the commits?

mfeurer:
No, thanks. I'll use the github functionality to do so if that's okay with you.

Issue #110, opened 2019-02-12
Errors in read/write of Config Space with Forbidden and Condition clauses


Neeratyoy:
Hello,

This is similar to the open issue #79.
Some more observations.
Initially had trouble writing to JSON, however, with the latest release that looks fine. 
I am still unable to write to PCS using pcs_new.write getting an error:
TypeError: sequence item 0: expected str instance, int found

Also, I tried to run the source code manually and observed the following:
[amsgrad | model_optimizer == 'adam',
 momentum | model_optimizer == 'sgd',
 channel_2 | n_conv_layer in {2, 3},
 channel_3 | n_conv_layer == 3,
 kernel_2 | n_conv_layer in {2, 3},
 kernel_3 | n_conv_layer == 3,
 padding_2 | n_conv_layer in {2, 3},
 padding_3 | n_conv_layer == 3,
 stride_2 | n_conv_layer in {2, 3},
 stride_3 | n_conv_layer == 3,
 (maxpool_2 | n_conv_layer in {2, 3} && maxpool_2 | stride_2 != 2),
 (maxpool_3 | n_conv_layer in {2, 3} && maxpool_3 | stride_3 != 2),
 (maxpool_kernel_2 | n_conv_layer in {2, 3} && maxpool_kernel_2 | maxpool_2 == 'True'),
 (maxpool_kernel_3 | n_conv_layer in {2, 3} && maxpool_kernel_3 | maxpool_3 == 'True'),
 maxpool_1 | stride_1 != 2,
 maxpool_kernel_1 | maxpool_1 == 'True']

On running the forbidden clauses with IN, such as
`(maxpool_3 | n_conv_layer in {2, 3} && maxpool_3 | stride_3 != 2)`
`(maxpool_kernel_2 | n_conv_layer in {2, 3} && maxpool_kernel_2 | maxpool_2 == 'True')`
 with _condition_lines.write(build_conjunction(condition))_ gives the following error:

> TypeError                                 Traceback (most recent call last)
> <ipython-input-347-da6d90edc711> in <module>()
> ----> 1 condition_lines.write(build_condition(condition))
> <ipython-input-284-cef9d614e16d> in build_condition(condition)
>     133         return in_template % (condition.child.name,
>     134                               condition.parent.name,
> --> 135                               ", ".join(condition.values))
>     136 
>     137     elif isinstance(condition, EqualsCondition):
> TypeError: sequence item 0: expected str instance, int found

On running the conditional clauses with IN, such as
[...]

Issue #79, opened 2018-06-12
TypeError; Write Conditions with non-string values


mlindauer:
```
from ConfigSpace.configuration_space import ConfigurationSpace
from ConfigSpace.hyperparameters import CategoricalHyperparameter
from ConfigSpace.conditions import InCondition
from ConfigSpace.read_and_write.pcs_new import write

cs = ConfigurationSpace(seed=12345)

p1 = CategoricalHyperparameter(name="p1", choices=[True, False], default_value=True)
p2 = CategoricalHyperparameter(name="p2", choices=[True, False], default_value=True)
cs.add_hyperparameters([p1,p2])
c1 = InCondition(child=p2, parent=p1, values=[True])
cs.add_condition(c1)

with open("test.pcs", "w") as fh:
    fh.write(write(cs))
```

Error:
```
Traceback (most recent call last):
  File "test.py", line 16, in <module>
    fh.write(write(cs))
  File "/home/lindauer/git/ConfigSpace/ConfigSpace/read_and_write/pcs_new.py", line 462, in write
    condition_lines.write(build_condition(condition))
  File "/home/lindauer/git/ConfigSpace/ConfigSpace/read_and_write/pcs_new.py", line 145, in build_condition
    ", ".join(condition.values))
TypeError: sequence item 0: expected str instance, bool found

```

In response to https://github.com/automl/SMAC3/pull/442

PR #247, opened 2022-04-11
check type of parent of conditions before writting to PCS files

dengdifan:
Should solve pcs write related issues: https://github.com/automl/ConfigSpace/issues/79 and https://github.com/automl/ConfigSpace/issues/110

### DIFF
--- a/ConfigSpace/read_and_write/pcs_new.py
+++ b/ConfigSpace/read_and_write/pcs_new.py
@@ -169,11 +169,10 @@ def build_condition(condition):
     notequal_template = "%s | %s != %s"
     equal_template = "%s | %s == %s"
 
-    parent = condition.parent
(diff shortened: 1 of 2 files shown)


ConfigSpace/read_and_write/pcs_new.py
```diff
@@ -169,28 +169,34 @@ def build_condition(condition):
     notequal_template = "%s | %s != %s"
     equal_template = "%s | %s == %s"
 
+    parent = condition.parent
+    if isinstance(parent, NumericalHyperparameter):
+        cond_values = str(condition.value)
+    else:
+        cond_values = [str(value) for value in condition.value]
```
mfeurer, line 176: This looks like it is not correct. There could be an `InCondition` on a UniformIntegerHyperparameter and it would not be treated correctly.

PR #255, opened 2022-06-14
Easy api

eddiebergman:
This PR aims to
1) Add some functionality to make creating config spaces less verbose
2) Maintain a pure python API layer for creation
3) Unify disparities between constructions of types of parameters.
4) Update documentation
5) Not break any old code

---
To facilitate automl/SMAC3#847, in which the user needs to know about ConfigSpace before they can even try SMAC, this PR allows basic ConfigSpace's to be built using a dict

```python
cs = ConfigurationSpace({
    "a": (0.1, 1.5),  # UniformFloat
    "b": (2, 10),  # UniformInt
    "c": ["cat", "dog", "mouse"]  #Categorical
})
```
* Only the above three are simple to support and rather than complicate this and parsing logic, users could use the next API points to "progress" from the easiest use case
* Using set vs list to distinguish between `Categorical` and `Ordinal` wasn't possible as Categorical's state that they are indeterminate w.r.t. to having a set. This might be possible to solve by simply sorting the received set in `Categorical` but outside of scope.
---
Another thing in this PR is to provide a simpler interface for constructing hyperparameters.
1. Unifies their construction
2. Less verbose
3. Leaves open the distribution to be separate from the hyperparameter type

```python
from ConfigSpace import Integer, Float, Categorical, Uniform, Normal, Beta

# Similar arguments as you're used to
x = Integer("a", (2, 20))
x = Integer("a", (2, 2000), log=True, quantization=3, default=3)

# Almost identical for Float ...
x = Float("a", (2.34, 2000), log=True, quantization=3, default=3)

a = Categorical("a", ["cat", "dog", "mouse"])
b = Categorical("b", ["small", "medium", "large"], ordered=True)  # Use the ordered flag

# Distributions
a = Integer("a", (1, 10), distribution=Uniform()) 
b = Integer("b", distribution=Normal(1, 10))  # Could also specify bounds
c = Integer("c", distribution=Beta(1, 10))
```

These are type safe so Mypy correctly knows that `b` is an `OrdinalHyperparamter` here, as well as for the rest of them.

---
The two of these combine to form:
```python
from ConfigSpace import ConfigurationSpace

cs = ConfigurationSpace({
    # Simple stuff stays simple
    "edge": ["sigmoid", "relu", ...],
    "depth": (1, 10),
    
    # Level up, you need more
    "lr": Float("lr", (0.1, 1), log=True, default=0.2),
    "other": Category("other", ["a", "b", "c"], weights=[1, 5, 1]),
    
    # Still use old ones too
    "old_param": UniformIntegerHyperparamter("old_param", ...)
})

# Or as keyword arguments
cs = ConfigurationSpace(
    seed=1234,
    space={ ... }
)
```

This also updates all documentation to reflect this (as well as update documentation in general)

You can run the following to download, build and view the docs:
```bash
git clone "https://github.com/automl/ConfigSpace" && cd ConfigSpace && git checkout easy_api && python -m venv .config_space_venv_tmp && source .config_space_venv_tmp/bin/activate && pip install -e ".[dev]" && make docs && rm -rf .config_space_venv_tmp && deactivate && python -c "import webbrowser; webbrowser.open('docs/build/html/index.html')"
```

Note that it has to unfortunately build the package first because Cython (grrr) needs to read the docstrings and generate something? That's why it's so long but the command mostly cleans up after itself, you just have to delete the `ConfigSpace` folder created once you're done. Might this might not work on Windows/Mac


renesass:
I really like how you can interact with configspace right now. I added examples in both readme and docs.
Good work, Edds!

eddiebergman:
Waiting to rebase onto #259 to fix errors in tests

eddiebergman:
I had a look, I like the deflated version of the API quite a bit! I think this is ready for merge and release at this point.

renesass:
Let me fix the comment thing tomorrow and then we can merge. :)

eddiebergman:
@mfeurer If you want to have a look at the new docs, they were cleaned up a lot and reworked to show off the simpler interface more.

```bash
git clone "https://github.com/automl/ConfigSpace" && cd ConfigSpace && git checkout easy_api && make doc && python -c "import webbrowser; webbrowser.open('docs/build/html/index.html')" 
```

ConfigSpace/conditions.pyx
```diff
@@ -169,26 +168,21 @@ cdef class EqualsCondition(AbstractCondition):
...
 
-        make *b* an active hyperparameter if *a* has the value 1
+            # make *b* an active hyperparameter if *a* has the value 1
+            cond = EqualsCondition(b, a, 1)
+            cs.add_condition(cond)
 
-        >>> cond = CS.EqualsCondition(b, a, 1)
-        >>> cs.add_condition(cond)
-        b | a == 1
+            # b | a == 1
```
mfeurer, line 185: The previous formatting would run doctest on the examples. Does the new formatting do the same?

ConfigSpace/api/__init__.py
```diff
@@ -0,0 +1,14 @@
+import ConfigSpace.api.distributions as distributions
+from ConfigSpace.api.distributions import Beta, Distribution, Normal, Uniform
+from ConfigSpace.api.types import Categorical, Float, Int
+
+__all__ = [
+    "distributions",
```
mfeurer, line 6: Hey, why do you advertise `"distributions"` here and not `"types"`? Or asked differently, is there a reason to advertise `"distributions"` in the first place? I'd be in favor of dropping it here and hiding the internal complexity.

ConfigSpace/configuration_space.pyx
```diff
@@ -1507,11 +1583,11 @@ class ConfigurationSpace(collections.abc.Mapping):
                     substituted_forbidden = forbidden_type(hyperparameter=new_hyperparameter, values=forbidden_arg)
                 else:
                     raise AttributeError(f'Did not find the expected attribute in forbidden {type(forbidden)}.')
-                
+
                 new_forbiddens.append(substituted_forbidden)
             else:
                 raise TypeError(f'Did not expect the supplied forbidden type {type(forbidden)}.')
-        
+
```
mfeurer, line 1590: Great to see all these formatting improvements - did you manage to set up the tooling for Cython files?

ConfigSpace/forbidden.pyx
```diff
@@ -218,25 +218,19 @@ cdef class MultipleValueForbiddenClause(AbstractForbiddenClause):
...
 
-    Example
-    -------
+    .. code:: python
 
-    >>> cs = CS.ConfigurationSpace()
-    >>> a = CSH.CategoricalHyperparameter('a', [1,2,3])
-    >>> cs.add_hyperparameters([a])
-    [a, Type: Categorical, Choices: {1, 2, 3}, Default: 1]
+        cs = ConfigurationSpace({"a": [1, 2, 3]})
```
mfeurer, line 228: I really like how much shorter the examples are now.

ConfigSpace/forbidden.pyx
```diff
@@ -218,25 +218,19 @@ cdef class MultipleValueForbiddenClause(AbstractForbiddenClause):
...
 
-    It forbids the value 2 for the hyperparameter *a*
+        # forbids the value 2 for the hyperparameter *a*
+        forbidden_clause_a = ForbiddenEqualsClause(cs["a"], 2)
 
-    >>> forbidden_clause_a = CS.ForbiddenEqualsClause(a, 2)
-    >>> cs.add_forbidden_clause(forbidden_clause_a)
-    Forbidden: a == 2
+        cs.add_forbidden_clause(forbidden_clause_a)
```
mfeurer, line 233: Following up on my questions from above: the doctest is removed from here, is that on purpose?

docs/api/hyperparameters.rst
```diff
@@ -0,0 +1,106 @@
...
+===============
+ConfigSpace contains
+:func:`~ConfigSpace.api.types.float.Float`,
+:func:`~ConfigSpace.api.types.int.Int`
+and :func:`~ConfigSpace.api.types.categorical.Categorical` hyperparamters, each with their own customizability.
+
+For :func:`~ConfigSpace.api.types.float.Float` and :func:`~ConfigSpace.api.types.int.Int`, you will find their
+interface much the same, being able to take the same :ref:`distributions <Distributions>` while :func:`~ConfigSpace.api.types.categorical.Categorical` can take weights or be ordered.
+
+These are all convenience functions that construct the more complex :ref:`hyperparameter classes <Advanced_Hyperparameters>` which make up the backbone of what's possible.
```
mfeurer, line 13: The header now drops the mention of ordinal hyperparameters, is that on purpose? Also, maybe it would be good to briefly mention that there are simple and advanced hyperparameter classes, and how they differ.

examples/README.rst
```diff
@@ -0,0 +1,4 @@
+Examples
+--------
+
+testtest
```
mfeurer, line 4: This looks unfinished?

examples/test.py
```diff
@@ -0,0 +1,8 @@
+"""
```
mfeurer, line 1: This looks unfinished, too.

mfeurer (commented):
Hey folks, I started having a look at this PR, but I have a few questions before I can continue looking further. In any case, it looks very nice and I like the new API very much.

ConfigSpace/conditions.pyx
eddiebergman: No, there's no doctest run. This was basically done to remove the `>>>` in the generated docs and try to make it as clean looking as possible. This does possibly introduce some inaccuracy between the docs and the output a user might receive if we do functional changes.

I'm okay with putting them back in but it does make the docs a little more "crowded" looking for the simple examples.

ConfigSpace/api/__init__.py
eddiebergman: The main reason was to allow for users to do something like the following incase `Normal` is already in the namespace
```python

class Normal:
   """User defined class or function"""

from ConfigSpace.api import distributions

Int("a", distribution=distributions.Normal(2, 3))
```

eddiebergman: But you raise a valid point that if we do that, we should also probably export `types` in a similar fashion.

I would be against dropping it but I'm really not pushed either way.

ConfigSpace/configuration_space.pyx
eddiebergman: No I didn't do anything on that front, I just ran a command I have to `StripTrailingWhitespace` that I usually use on files that have it

ConfigSpace/forbidden.pyx
eddiebergman: :)

ConfigSpace/forbidden.pyx
eddiebergman: See above

docs/api/hyperparameters.rst
eddiebergman: So I guess that is capture in mentioning "...`Categorical` can take weights or be **ordered**" but maybe we should emphasise this more.

Will do on the second point on mentioning how they differ.

examples/test.py
eddiebergman: Also needs to be removed, I think this is an artifact of setting up the sphinx-automl-theme

renesass: We could also move the user guide to the examples? Like here https://automl.github.io/automl_sphinx_theme/main/examples/index.html

ConfigSpace/conditions.pyx
renesass: How about we just move those tests to the tests directory?

renesass:
One thing which came up earlier in my mind was that we have 
- Int
- Float
- Categorical

Int, in contrast to float and categorical, is shortcut. Should we change it to Integer for consistency?

examples/test.py
eddiebergman: If you're willing to move each actual examples in the user guide to their own examples then go for it, otherwise I would advocate to just delete the need for examples. Would this break the theme?

ConfigSpace/conditions.pyx
eddiebergman: Those things are already tested (I hope)

mfeurer:
> Int, in contrast to float and categorical, is shortcut. Should we change it to Integer for consistency?

Sound good.

eddiebergman:
Well I don't think `Float` has a well known shortcut and `Cat` is a different word but sure, I don't really mind. I was going with programming syntax words.

mfeurer:
I had an offline discussion with @eddiebergman and we agreed to add the doctests back in. I will do a full review after these changes.

eddiebergman:
@mfeurer I've updated to doctests again. Took a long ass while to debug why things weren't working but `make doc` now correctly does everything. We have to rebuild the package for the docstrings to properly be exported and doctested which is annoying but I don't see a way around it.

I've also included the changelog in this PR since it bumps the version number, theoretically this could be done in a seperate PR but I wanted to just get it done in one place.

Seems there is also some tolerance needed for a test `AssertEqual -> AssertAlmostEqual` to get bumped for one test (15th decimal place). Again, could be done as a seperate PR but for expediency, its included in this.

@renesass just a note when changing from `master` to `main` which you might do elsewhere, make sure to check the docs and workflow files for references to master and change them as well.

setup.py
```diff
@@ -9,103 +9,118 @@
...
 )
 KEYWORDS = (
-    'algorithm configuration hyperparameter optimization empirical '
-    'evaluation black box'
+    "algorithm configuration hyperparameter optimization empirical "
+    "evaluation black box"
+)
+LICENSE = "BSD 3-clause"
+PLATS = ["Linux"]
+AUTHORS = (
```
KEggensperger, line 47: Could you use the same list as in __init__ (or even import, read it from there)

KEggensperger (approved):

setup.py
```diff
@@ -9,103 +9,118 @@
...
+    "automated algorithm configuration and hyperparameter tuning."
 )
 KEYWORDS = (
-    'algorithm configuration hyperparameter optimization empirical '
-    'evaluation black box'
+    "algorithm configuration hyperparameter optimization empirical "
+    "evaluation black box"
+)
+LICENSE = "BSD 3-clause"
+PLATS = ["Linux"]
```
mfeurer, line 46: Didn't you extend this to also work on Windows and OSX?

ConfigSpace/api/__init__.py
mfeurer: Could you then please export `types`?

mfeurer (approved):
Very minor, looks good to go :)

eddiebergman:
Extra few things done, only possible slowdown is the "docs" workflow, I hope that doesn't catch. There was a numerical thing to do with some floating point value and I hope it doesn't trigger

PR #256, opened 2022-06-17
Fix small bugs related to nan values in vector passed to pdf

Ennosigaeon:
I tried to incorporate the new prior system with SMAC and noticed some bugs regarding NaN values introduced by conditional hyperparameters. This PR resolves all issues I have encountered with conditional hyperparameters.

If you would like to, I can also provide a minimal working example showcasing the current problems.

### DIFF
--- a/ConfigSpace/hyperparameters.pyx
+++ b/ConfigSpace/hyperparameters.pyx
@@ -2609,7 +2609,13 @@ cdef class CategoricalHyperparameter(Hyperparameter):
             Probability density values of the input vector
         """
         probs = np.array(self.probabilities)
+        nan = np.isnan(vector)
+        if np.any(nan):
+            # Temporarily pick any valid index to use `vector` as an index for `probs`
+            vector[nan] = 0
         res = np.array(probs[vector.astype(int)])
+        if np.any(nan):
+            res[nan] = 0
         if res.ndim == 0:
             return res.reshape(-1)
         return res
--- a/test/test_configuration_space.py
+++ b/test/test_configuration_space.py
@@ -46,7 +46,7 @@ from ConfigSpace.hyperparameters import (UniformFloatHyperparameter,
                                          BetaIntegerHyperparameter,
                                          OrdinalHyperparameter)
 from ConfigSpace.exceptions import ForbiddenValueError
-from ConfigSpace.forbidden import ForbiddenEqualsRelation
+from ConfigSpace.forbidden import ForbiddenEqualsRelation, ForbiddenLessThanRelation
 
 
 def byteify(input):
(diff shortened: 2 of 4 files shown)


Ennosigaeon:
Hey, I just found some time to finally rework this PR. There are now two changes, each in a separate commit and with according unit tests.

1. It is currently not possible to call ConfigSpace.substitute_hyperparameters_in_* with either InConditions or ForbiddenRelations. The first commit resolves this issue.
2. If you are using ConfigSpaces with conditional hyperparameters, some config values in the array representation may be nan. In the current SMAC implementation of [user priors](https://github.com/automl/SMAC3/blob/83a9bbe57e64968ee0e79d5bbedf21a1f835719e/smac/optimizer/acquisition/__init__.py#L276)  all values, including potential nans, are passed to _pdf. This commit returns a PDF of 0 for nan values.

Regarding the minimal working example: You can use the changed unit tests to provoke exceptions with the current main.

eddiebergman:
Amazing, thanks for the update, I will check this once I'm back next week!

eddiebergman (approved):
So it looks good to me. I think @mfeurer might be able to review the first part regarding the InCondition and ForibiddenClauses but the second part about the nans make sense.

Am I correct to assume that your changes for the clauses stems from the fact the `elif` statement was never reached as they always contained the attribute `"value"`?

Ennosigaeon:
Yes, your assumption is correct. An additional `values` field is introduced but `value` also exists.

ConfigSpace/hyperparameters.pyx
```diff
@@ -2556,7 +2556,14 @@ cdef class CategoricalHyperparameter(Hyperparameter):
             Probability density values of the input vector
         """
         probs = np.array(self.probabilities)
+        vector = np.atleast_1d(vector)
```
mfeurer, line 2559: Hey, is there a reason to add this only for categorical hyperparameters?

ConfigSpace/hyperparameters.pyx
```diff
@@ -2556,7 +2556,14 @@ cdef class CategoricalHyperparameter(Hyperparameter):
...
+        nan = np.isnan(vector)
+        if np.any(nan):
+            vector[nan] = 0
         res = np.array(probs[vector.astype(int)])
+        if np.any(nan):
+            res[nan] = 0
+            vector[nan] = np.nan
```
mfeurer, line 2566: This is not used any more and appears to be not necessary.

ConfigSpace/hyperparameters.pyx
```diff
@@ -2556,7 +2556,14 @@ cdef class CategoricalHyperparameter(Hyperparameter):
             Probability density values of the input vector
         """
         probs = np.array(self.probabilities)
+        vector = np.atleast_1d(vector)
+        nan = np.isnan(vector)
+        if np.any(nan):
+            vector[nan] = 0
```
mfeurer, line 2562: As the res[nan] is set to zero further belows, this appears if-block appears to be unnecessary.

test/test_configuration_space.py
```diff
@@ -919,6 +919,27 @@ def test_substitute_hyperparameters_in_conditions(self):
...
+        sub_a = UniformIntegerHyperparameter('a', lower=0, upper=10)
+        sub_b = UniformFloatHyperparameter('b', lower=1., upper=8., log=False)
+        cs2.add_hyperparameters([sub_a, sub_b])
+        new_conditions = cs1.substitute_hyperparameters_in_conditions(cs1.get_conditions(), cs2)
+
+        test_cond = InCondition(b, a, [1, 2, 3, 4])
+        cs2.add_conditions([test_cond])
+        test_conditions = cs2.get_conditions()
+
+        self.assertEqual(new_conditions[0], test_conditions[0])
```
mfeurer, line 941: Can you make this test stricter? For example by checking that the two conditions are not the same instance of the object?

mfeurer (commented):
Hi folks, sorry for being slow to catch up after ICML. Finally, I had time to look at your PR and have left some further remarks.

mfeurer:
Hey @Ennosigaeon, any updates on this? It would be great if we could merge this PR as the number of merge issues is increasing.

mfeurer:
Hey, sorry to bring this up again, but we're starting to move hyperparameters into individual files in #292. It would be great if we could merge this one before to avoid extensive merge conflicts..

Ennosigaeon:
I am on vacation this week. I will take a look at it beginning of next week. Promised ;)

ConfigSpace/hyperparameters.pyx
Ennosigaeon: The problem is that vector is used in the next line to access the probabilities from `probs`. We need any valid index for `probs`. Therefore, I just picked `0`. I added a comment to make this clear

test/test_configuration_space.py
Ennosigaeon: I am not sure what to exactly test here. I have added a check for same instance, equals parent HP, same instance parent HP, equal child HP and same instance child HP. I hope that's sufficient.

ConfigSpace/hyperparameters.pyx
Ennosigaeon: Removed line 2566

ConfigSpace/hyperparameters.pyx
Ennosigaeon: Not sure why I added this, but everything is still working without this line. I removed it.

Ennosigaeon:
Ok, everything looks fine from my side. I will just ignore the failed tests due to `The job was not started because recent account payments have failed or your spending limit needs to be increased. Please check the 'Billing & plans' section in your settings.`

Please let me know if you want me to change anything else.

test/test_configuration_space.py
```diff
@@ -919,6 +919,33 @@ def test_substitute_hyperparameters_in_conditions(self):
...
+        self.assertEqual(new_conditions[0].get_parents(), test_conditions[0].get_parents())
+        self.assertFalse(new_conditions[0].get_parents() is test_conditions[0].get_parents())
```
mfeurer, line 943:
```suggestion
        self.assertIs(new_conditions[0].get_parents(), test_conditions[0].get_parents())
```

test/test_configuration_space.py
mfeurer: I meant an additional `self.assertIsNot(new_conditions[0], test_conditions[0])` which checks that a new object was created.

mfeurer:
Thanks a lot for these updates. Indeed, github appears to have hickups and this is not due to this PR.

test/test_configuration_space.py
Ennosigaeon: I think it should be `assertIsNot`. I changed it accordingly

mfeurer:
Thanks a lot. Github has resolved the hickups, so I just merged your contribution.

PR #268, opened 2022-08-23
short part

eddiebergman:
If you need a longer explatnation of stuff, here's a good way to do it
and it will show up not in the short part

### DIFF
--- a/myfile.txt
+++ b/myfile.txt
@@ -1,0 +1,1 @@
+changes


myfile.txt
```diff
@@ -0,0 +1 @@
+changes
```
eddiebergman, line 1: hey this should just be `change`

eddiebergman (commented):
Overall looks good, some minor changes needed

Issue #277, opened 2022-11-15
An implementation to support complex conditions such as 'a < b and a * b < 10 and ...'


jhj0411jhj:
I implement a `ConditionedConfigurationSpace` that supports complex conditions between hyperparameters (e.g., x1 <= x2 and x1 * x2 < 100). User can define a `sample_condition` function to restrict the generation of configurations.

The following functions are guaranteed to return valid configurations:
+ self.sample_configuration()
+ get_one_exchange_neighbourhood()  # may return empty list

Here is an example:
```python
def sample_condition(config):
    # require x1 <= x2 and x1 * x2 < 100
    if config['x1'] > config['x2']:
        return False
    if config['x1'] * config['x2'] >= 100:
        return False
    return True
    # return config['x1'] <= config['x2'] and config['x1'] * config['x2'] < 100

cs = ConditionedConfigurationSpace()
cs.add_hyperparameters([...])
cs.set_sample_condition(sample_condition)  # set the sample condition after all hyperparameters are added
configs = cs.sample_configuration(1000)
```

Implementing this feature using fobiddens like `ForbiddenClause` or `ForbiddenRelation` might be a viable option, but it's a little complicated, and user may need to pass the full config space to the forbidden object.

The implementation does not consider serialization (of sample_condition function).

Here is the implementation of `ConditionedConfigurationSpace`:
```python
from typing import List, Union, Callable
import numpy as np
from ConfigSpace import ConfigurationSpace, Configuration
from ConfigSpace.exceptions import ForbiddenValueError


class ConditionedConfigurationSpace(ConfigurationSpace):
    """
    A configuration space that supports complex conditions between hyperparameters,
        e.g., x1 <= x2 and x1 * x2 < 100.

    User can define a sample_condition function to restrict the generation of configurations.

    The following functions are guaranteed to return valid configurations:
        - self.sample_configuration()
        - get_one_exchange_neighbourhood()  # may return empty list

    Example
    -------

    >>> def sample_condition(config):
    >>>     # require x1 <= x2 and x1 * x2 < 100
    >>>     if config['x1'] > config['x2']:
    >>>         return False
    >>>     if config['x1'] * config['x2'] >= 100:
    >>>         return False
    >>>     return True
    >>>
    >>> cs = ConditionedConfigurationSpace()
    >>> cs.add_hyperparameters([...])
    >>> cs.set_sample_condition(sample_condition)  # set the sample condition after all hyperparameters are added
[...]

Issue #272, opened 2022-09-24
Allowing condition such as 'a<b'?


oleotiger:
There is a widespread situation in hyperparameter tuning/knob tuning problem. A hyperparameter $a$ must be less than another parameter $b$. Or more complex, $a<b<c$. 

Is it possible to support such condition in 'ConfigSpace'?

mfeurer:
I'm afraid no, and we only have this for [forbidden clauses](https://github.com/automl/ConfigSpace/blob/main/ConfigSpace/forbidden.pyx#L487). Such a contribution would be very welcome, though.

jhj0411jhj:
I implement a `ConditionedConfigurationSpace` to support such complex conditions. Please refer to #277 for more details.

Issue #254, opened 2022-05-31
How to create Conditions based on multiple variables likes a+b < c??


xdqkid:
How to realize the condition likes a+b < c?

thijssnelleman:
Closing this as its a duplicate now with #277 containing the most information

PR #280, opened 2022-12-05
Add ForbiddenCallableRelation

nchristensen:
Allows for more complicated forbidden relationships using Callables. Related issues #254 #272 #277


eddiebergman:
Hi @nchristensen,

While this is quite cool, and definitely something that's desirable, it makes somethings impossible which is something we definitely want to keep, namely serializibility. We can't save configspaces with callables to JSON.

Another example, close to ConfigSpace is [`ParameterSpace`](https://boschresearch.github.io/parameterspace/v0.7.14/) which does allow for the callables but lacks serializibility for general functions. They do serialize `lambda` functions but I can't speak to it's robustness.

In light of this, I don't want to outright reject the PR, it's still nice to have but I imagine it will require some explicit testing with respect to serialization and we need explicit documentation that use of this feature would prevent non-binary serialization to disk. (You could still pickle them of course but pickles are a bit volatile with long term storage).

I'll review this later today and drop some pointers as to where this could be done!

ConfigSpace/forbidden.pyx
```diff
@@ -564,13 +568,74 @@ cdef class ForbiddenRelation(AbstractForbiddenComponent):
...
+
+    The ForbiddenCallable uses two hyperparameters as input to a
+    specified callable, which returns True if the relationship
+    between the two hyperparameters is forbidden.
+
+    >>> from ConfigSpace import ConfigurationSpace, ForbiddenLessThanRelation
+    >>>
+    >>> cs = ConfigurationSpace({"a": [1, 2, 3], "b": [2, 5, 6]})
+    >>>
+    >>> forbidden_clause = ForbiddenCallableRelation(cs['a'], cs['b'])
```
eddiebergman, line 591: This example seems to be missing a callable `f` defining the relationship between the two?

ConfigSpace/forbidden.pyx
```diff
@@ -564,13 +568,74 @@ cdef class ForbiddenRelation(AbstractForbiddenComponent):
...
+
+    Parameters
+    ----------
+     left : :ref:`Hyperparameters`
+         first argument of callable
+
+     right : :ref:`Hyperparameters`
+         second argument of callable
+
+     f : A callable that relates the two hyperparameters
```
eddiebergman, line 603: Sphinx will complain about the documentation having a leading space on this line unfortunately, it's very particular :/

ConfigSpace/forbidden.pyx
```diff
@@ -564,13 +568,74 @@ cdef class ForbiddenRelation(AbstractForbiddenComponent):
...
+        if not isinstance(f, Callable):  # Can't use ForbiddenCallable here apparently
+            raise TypeError("Argument 'f' is not of type %s." % Callable)
+
+        super().__init__(left, right)
+        self.f = f
+
+    def __eq__(self, other: Any) -> bool:
+        if not isinstance(other, self.__class__):
+            return False
+        return super().__eq__(other) and self.f == other.f
```
eddiebergman, line 618: Nice use of the super class to keep consistency in equality checking.

ConfigSpace/forbidden.pyx
```diff
@@ -564,13 +568,74 @@ cdef class ForbiddenRelation(AbstractForbiddenComponent):
...
+        return self.__class__(
+            a=copy.copy(self.left),
+            b=copy.copy(self.right),
+            f=copy.copy(self.f)
+        )
+
+    def __repr__(self):
+        from inspect import getsource
+        f_source = getsource(self.f)
+        return f"Forbidden: {f_source} | Arguments: {self.left.name}, {self.right.name}"
```
eddiebergman, line 630: This `getsource` call could be quite large and rather verbose. For example:

```python
def f():
    print("hello")
    x = 1 + 2
    y = "hi" + "mars"
    return b"no"
    
from inspect import getsource
getsource(f)
# 'def f():\n    print("hello")\n    x = 1 + 2\n    y = "hi" + "mars"\n    return b"no"\n'
```

I think we could just get around this by using the functions, `f.__qualname__` instead.

ConfigSpace/forbidden.pyx
```diff
@@ -564,13 +568,74 @@ cdef class ForbiddenRelation(AbstractForbiddenComponent):
...
+
+    cdef int _is_forbidden(self, left, right) except -1:
```
eddiebergman, line 632: What does this `except -1` do? I wish I knew more Cython but unfortunatly not. I looked at the class above and it didn't seem to have this.

ConfigSpace/hyperparameters.pyx
```diff
@@ -1048,13 +1050,14 @@ cdef class NormalFloatHyperparameter(FloatHyperparameter):
     def _sample(self, rs: np.random.RandomState, size: Optional[int] = None
                 ) -> Union[np.ndarray, float]:
 
-        if self.lower == None:
+        sigma = self.sigma
+        if sigma == 0:
+            return self.mu
+        elif self.lower == None:
```
eddiebergman, line 1056: This is a nice change in of itself. Depending on how this PR goes, I still think we'd like to pull this in

ConfigSpace/hyperparameters.pyx
```diff
@@ -1710,6 +1714,8 @@ cdef class NormalIntegerHyperparameter(IntegerHyperparameter):
 
         self.mu = mu
         self.sigma = sigma
+        if self.sigma == 0:
+            assert isinstance(self.mu, int)
```
eddiebergman, line 1718: Why this assertion? I feel like they should both just be `int`. The init signature seems off here to suggest otherwise.

ConfigSpace/forbidden.pyx
eddiebergman: As a follow up you could accomodate lambdas as:

```python
qualname = f.__qualname__
f_repr = getsource(f) if qualname == "<lambda>" else qualname
```

ConfigSpace/hyperparameters.pyx
```diff
@@ -1663,8 +1668,7 @@ cdef class NormalIntegerHyperparameter(IntegerHyperparameter):
     cdef public nfhp
     cdef normalization_constant
 
-
-    def __init__(self, name: str, mu: int, sigma: Union[int, float],
+    def __init__(self, name: str, mu: Union[int, float], sigma: Union[int, float],
```
eddiebergman, line 1671: Seems you explicitly type allow `float`? How come?

eddiebergman (commented):
So functionally, it all seems there but as mentioned in the PR comment, we definitely need tests to check its behaviour when conisdering serialization.

My guess is serialization will fail and give some JSON based error. If this is the case. To counteract this, I think we should raise an explicit error stating that using a `ForbiddenCallable` means you can not serialize the space to JSON. This should also be stated in the documentation directly.

Thanks for the PR though, sorry for my critical comments. We do appreciate it, really and I'm sorry we don't dedicate more time to directly improving this library and user contributions are a nice surprise!

ConfigSpace/hyperparameters.pyx
nchristensen: My thinking is that even if we're drawing integers from a normal distribution, there is no requirement that the mean of the distribution be an integer. For instance, a user might want 4 and 5 to be chosen with equal probability and so place the mean at 4.5.

ConfigSpace/hyperparameters.pyx
nchristensen: This follows from below. But I'm also fine with restoring the existing behavior if allowing a non-integer mean is undesirable.

nchristensen:
Yeah, serializability could be difficult with this. I think it would be useful to optionally allow pickling of the Callable during serialization, but to disallow this by default.

ConfigSpace/forbidden.pyx
nchristensen: That's a good idea. I was noticing formatting issues using `getsource` as well since it keeps all of the leading white space.

ConfigSpace/forbidden.pyx
nchristensen: I modeled this off of the `ForbiddenLessThanRelation` directly below which uses this. I'm not very familiar with Cython either, but it seems any cdef function that might return a Python exception needs to be declared with an except value. https://docs.cython.org/en/latest/src/userguide/language_basics.html#error-return-values

ConfigSpace/forbidden.pyx
nchristensen: My knowledge of Sphinx is limited, but I revised it modeled on the ForbiddenEqualsRelation. Hopefully that resolves this.

ConfigSpace/forbidden.pyx
nchristensen: Obtaining only the source code of the lambda (and discarding any other code on the same line) is apparently non-trivial. See for example https://stackoverflow.com/questions/59498679/how-can-i-get-exactly-the-code-of-a-lambda-function-in-python. I have revised it to use the qualname though.

mfeurer:
Hey @nchristensen thank you very much for the PR. I'll discuss with @eddiebergman whether we will extend the ConfigSpace package to allow callables (no timeline, though). However, your changes to the hyperparameter file appear to be useful by themselves (as mentioned by @eddiebergman). Would you like to create a single PR so we can merge these anyway?

nchristensen:
Sure, it will probably be sometime in February before I can get back to this though.

PR #282, opened 2022-12-15
fix: Memory leak

### LINKED ISSUE #187: Memory Leak
By running the following code, the memory usage goes up like crazy:

```
import ConfigSpace.hyperparameters as CSH
import numpy as np
rnd = np.random.RandomState(19937)
a = CSH.UniformIntegerHyperparameter('a', lower=1, upper=2147483647, log=True)
print(a)
print(rnd)
for i in range(1,10000):
   a.get_neighbors(0.031249126501512327, rnd, number=8, std=0.05)
```

tested on kisexe44 with the current master branch (0.4.19).

eddiebergman:
* Closes #187

A core issue addressed was the following setup:
1. `get_neighbors` with a small std. deviation.
2. A small set of neighbors to choose from.

This resulted in rejection sampling never finding enough valid neighbors in its Gaussian from which to sample.

---
The solution was to simply find the next closest neighbor that was not sampled from yet. Iteratively checking around that number until one was found. This also respects the boundaries.

```python
neighbors = [3, 4, 5, 6]
sample = 5
# try 4, nope
# try 6, nope
# try 3, nope
# try 7, yup
neighbors.append(7)
```

I'm not sure if this is "correct" but I don't see another viable solution.

---
Another issue came up, as seen in the comment below in that the `NormalIntegerHyperparamater` (and later diagnosed to also be a problem for `BetaIntegerHyperparameter`) is that the `_compute_normalization` and `get_max_density` require computing the pdf for every possible int value. This caused memory to blow up in the case where the range of possible values was incredibly large, for example every possible int32 value.

To combat this, I implemented an `arange_chunked` which functions similarly to `arange` but yields sub-chunks. This was possible because the `sum` and `max` operations are possible over partial chunks and do not require the full `arange` to be in memory at once.

This is still incredibly slow, calculating the `pdf` and `max` density over this entire range and it's likely that an analytical solution is possible, as we deal with subsequent numbers.

This is documented in #283

---
It's also quite difficult to work with this codebase given the `.pyx` doesn't allow for editors to be very smart (for example jump-to-defintion). Using normal text search is also frustrating due to all classes sharing method names and being in one file. Just voicing again that converting this back to pure python and splitting up the files would make working with ConfigSpace easier. Any performance issues which originally motivated the switch are likely solvable within python itself as numpy can do the heavy lifting in C.

### DIFF
--- a/ConfigSpace/functional.py
+++ b/ConfigSpace/functional.py
@@ -48,7 +48,7 @@ def arange_chunked(
 ) -> Iterator[np.ndarray]:
     """Get np.arange in a chunked fashion.
 
-    >>> list(sequential_chunks(0, 10, 3))
+    >>> list(arange_chunked(0, 10, 3))
     [array([0, 1, 2]), array([3, 4, 5]), array([6, 7, 8]), array([9])]
 
     Parameters
--- a/ConfigSpace/functional.py
+++ b/ConfigSpace/functional.py
@@ -34,8 +34,8 @@ def center_range(
     Iterator[int]
     """
     assert low <= center <= high
-    below_center = range(center + step, high + 1, step)
-    above_center = range(center - step, low - 1, -step)
+    above_center = range(center + step, high + 1, step)
+    below_center = range(center - step, low - 1, -step)
     yield from roundrobin(below_center, above_center)
 
 
--- a/ConfigSpace/functional.py
+++ b/ConfigSpace/functional.py
@@ -1,37 +1,7 @@
-from itertools import cycle, islice
-from typing import Iterator, Iterable, TypeVar
+from typing import Iterator
 
-T = TypeVar("T")
-
-
-def roundrobin(*iterables: Iterable[T]) -> Iterator[T]:
-    """Iterate over several iterables in a roundrobin fashion.
-
-    https://docs.python.org/2/library/itertools.html#recipes
-
-    >>> list(roundrobin('ABC', 'D', 'EF'))
-    ['A', 'D', 'E', 'B', 'F', 'C']
-
-    Parameters
-    ----------
-    *iterables: Iterable[T]
-        The iterables to use
-
-    Returns
-    -------
-    Iterator[T]
-    """
-    # Recipe credited to George Sakkis
-    num_active = len(iterables)
-    nexts = cycle(iter(it).__next__ for it in iterables)
-    while num_active:
-        try:
-            for next in nexts:
-                yield next()
-        except StopIteration:
-            # Remove the iterator we just exhausted from the cycle.
-            num_active -= 1
-            nexts = cycle(islice(nexts, num_active))
+from more_itertools import roundrobin
+import numpy as np
 
 
 def center_range(
@@ -41,6 +11,9 @@ def center_range(
     step: int = 1,
 ) -> Iterator[int]:
     """Get a range centered around a value.
+
+    >>> list(center_range(5, 0, 10))
+    [4, 6, 3, 7, 2, 8, 1, 9, 0, 10]
 
     Parameters
     ----------
@@ -64,3 +37,45 @@ def center_range(
     below_center = range(center + step, high + 1, step)
     above_center = range(center - step, low - 1, -step)
     yield from roundrobin(below_center, above_center)
+
+
+def arange_chunked(
+    start: int,
+    stop: int,
+    step: int = 1,
+    *,
+    chunk_size: int,
+) -> Iterator[np.ndarray]:
+    """Get np.arange in a chunked fashion.
+
+    >>> list(sequential_chunks(0, 10, 3))
+    [array([0, 1, 2]), array([3, 4, 5]), array([6, 7, 8]), array([9])]
+
+    Parameters
+    ----------
+    start: int
+        The start of the range
+
+    stop: int
+        The stop of the range
+
+    chunk_size: int
+        The size of the chunks
+
+    step: int = 1
+        The step size
+
+    Returns
+    -------
+    Iterator[np.ndarray]
+    """
+    assert step > 0
+    assert chunk_size > 0
+    assert start < stop
+    n_items = int(np.ceil((stop - start) / step))
+    n_chunks = int(np.ceil(n_items / chunk_size))
+
+    for chunk in range(0, n_chunks):
+        chunk_start = start + (chunk * chunk_size)
+        chunk_stop = min(chunk_start + chunk_size, stop)
+        yield np.arange(chunk_start, chunk_stop, step)
--- a/ConfigSpace/functional.py
+++ b/ConfigSpace/functional.py
@@ -71,7 +71,7 @@ def arange_chunked(
     """
     assert step > 0
     assert chunk_size > 0
-    assert 0 <= start < stop
+    assert start < stop
     n_items = int(np.ceil((stop - start) / step))
     n_chunks = int(np.ceil(n_items / chunk_size))
 
--- a/ConfigSpace/functional.py
+++ b/ConfigSpace/functional.py
@@ -1,0 +1,66 @@
+from itertools import cycle, islice
+from typing import Iterator, Iterable, TypeVar
+
+T = TypeVar("T")
+
+
+def roundrobin(*iterables: Iterable[T]) -> Iterator[T]:
+    """Iterate over several iterables in a roundrobin fashion.
+
+    https://docs.python.org/2/library/itertools.html#recipes
+
+    >>> list(roundrobin('ABC', 'D', 'EF'))
+    ['A', 'D', 'E', 'B', 'F', 'C']
+
+    Parameters
+    ----------
+    *iterables: Iterable[T]
+        The iterables to use
+
+    Returns
+    -------
+    Iterator[T]
+    """
+    # Recipe credited to George Sakkis
+    num_active = len(iterables)
+    nexts = cycle(iter(it).__next__ for it in iterables)
+    while num_active:
+        try:
+            for next in nexts:
+                yield next()
+        except StopIteration:
+            # Remove the iterator we just exhausted from the cycle.
+            num_active -= 1
+            nexts = cycle(islice(nexts, num_active))
+
+
+def center_range(
+    center: int,
+    low: int,
+    high: int,
+    step: int = 1,
+) -> Iterator[int]:
+    """Get a range centered around a value.
+
+    Parameters
+    ----------
+    center: int
+        The center of the range
+
+    low: int
+        The low end of the range
+
+    high: int
+        The high end of the range
+
+    step: int = 1
+        The step size
+
+    Returns
+    -------
+    Iterator[int]
+    """
+    assert low <= center <= high
+    below_center = range(center + step, high + 1, step)
+    above_center = range(center - step, low - 1, -step)
+    yield from roundrobin(below_center, above_center)
--- a/ConfigSpace/functional.py
+++ b/ConfigSpace/functional.py
@@ -71,7 +71,7 @@ def arange_chunked(
     """
     assert step > 0
     assert chunk_size > 0
-    assert start < stop
+    assert 0 <= start < stop
     n_items = int(np.ceil((stop - start) / step))
     n_chunks = int(np.ceil(n_items / chunk_size))
 
--- a/ConfigSpace/hyperparameters.pyx
+++ b/ConfigSpace/hyperparameters.pyx
@@ -34,8 +34,10 @@ from collections import OrderedDict, Counter
 from typing import List, Any, Dict, Union, Set, Tuple, Optional, Sequence
 
 from scipy.stats import truncnorm, beta as spbeta, norm
+import numpy as np
 cimport numpy as np
 
+from ConfigSpace.functional import center_range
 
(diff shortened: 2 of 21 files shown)


eddiebergman:
```python
import ConfigSpace.hyperparameters as CSH
import numpy as np
rnd = np.random.RandomState(19937)

# This gets to the for loop before hanging
#a = CSH.UniformIntegerHyperparameter('a', lower=1, upper=2147483647, log=True)

# This hangs before the prints
a = CSH.NormalIntegerHyperparameter('a', mu=10, sigma=500, lower=1, upper=2147483647, log=True)

print(a, flush=True)
print(rnd, flush=True)

for i in range(1, 10000):
    a.get_neighbors(0.031249126501512327, rnd, number=8, std=0.05)
```

eddiebergman:
Back to the original memory overflow:

```python
number = 5  # slow but fine
for i in range(1, 10000):
    a.get_neighbors(0.031249126501512327, rnd, number=number, std=0.05)
    
number = 6 # Suddenly blows up memory and unkillable process
for i in range(1, 10000):
    a.get_neighbors(0.031249126501512327, rnd, number=number, std=0.05)
```

ConfigSpace/hyperparameters.pyx
```diff
@@ -1882,31 +1905,46 @@ cdef class NormalIntegerHyperparameter(IntegerHyperparameter):
...
-
-                if self.lower is not None and self.upper is not None:
-                    int_value = min(max(int_value, self.lower), self.upper)
-                    new_int_value = min(max(new_int_value, self.lower), self.upper)
-
-                if int_value != new_int_value:
-                    rejected = False
-                elif iteration > 100000:
-                    raise ValueError('Probably caught in an infinite loop.')
+            possible_neighbor = self._sample(rs)
```
mfeurer, line 1916: This does no longer sample around the current value and thus thereby change the behavior of SMAC etc. Could you please change this? Maybe we can use a similar logic as above (except that we replace the truncated normal by a regular normal distribution)?

ConfigSpace/hyperparameters.pyx
```diff
@@ -1542,84 +1554,95 @@ cdef class UniformIntegerHyperparameter(IntegerHyperparameter):
 
     def get_neighbors(
```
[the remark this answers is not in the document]
eddiebergman: I thought I would just deal with memory issues for now, it's quite possible we could unite the neighbor generating algorithm into one lean function and not have many similar implementations.

ConfigSpace/hyperparameters.pyx
eddiebergman: Whoops, good catch, I had it like that before but realised we need different distributions depending on if it's `bounded` or not. Thought I could sneak it some re-use but this ignore `value` will fix.

ConfigSpace/hyperparameters.pyx
```diff
@@ -1542,84 +1554,95 @@ cdef class UniformIntegerHyperparameter(IntegerHyperparameter):
...
+                numbers_around = center_range(possible_neighbor, self.lower, self.upper, stepsize)
+
+                valid_numbers_around = (
+                    n for n in numbers_around
+                    if (n not in neighbors and n != center)
+                )
+                possible_neighbor = next(valid_numbers_around, None)
+
+                if possible_neighbor is None:
+                    raise ValueError(
```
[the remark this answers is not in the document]
eddiebergman: Essentially this highlighted ValueError is reached. Following from the logic:
1) Neighbors will attempt to be sampled from
2) As this neighbors set fills up, it's more and more likely to go into the `if possible_neighbor in neighbors` block.
3) This block will then create a generator `valid_numbers_around` which will generate all possible int values, centered at the `possible_neighbor`.
4) We iterate over this generator until we find one that's not included in the `neighbors` set.
5) If we exhaust the generator, we hit the `None` in `next(valid_numbers_around, None)` and trigger the error.

We could check this before hand and exit early but I figure it's more of an edge case that this happens and it seems the main user of this, `get_one_exchange_neighborhood` , checks the amount of neighbors first anyways.

ConfigSpace/hyperparameters.pyx
```diff
@@ -1542,84 +1554,95 @@ cdef class UniformIntegerHyperparameter(IntegerHyperparameter):
...
+            scale=std,
+        )
+
+        center = self._transform(value)
+
+        # Get random values from the truncated normal distribution and places
+        # them into neighbors
+        neighbors: set[int] = set()
+        while len(neighbors) < number:
+            float_index = hypercube_dist.rvs(random_state=rs)
```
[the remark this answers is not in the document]
eddiebergman: Hmm that's true, I'll give it a go

ConfigSpace/hyperparameters.pyx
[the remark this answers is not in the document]
mfeurer: Yes, I agree. This should be a separate PR.

ConfigSpace/functional.py
```diff
@@ -0,0 +1,81 @@
...
+        The size of the chunks
+
+    step: int = 1
+        The step size
+
+    Returns
+    -------
+    Iterator[np.ndarray]
+    """
+    assert step > 0
```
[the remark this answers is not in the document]
eddiebergman: Only positive step sizes are allowed but you could do an `arange(-10, 10, 1)` which is perfectly valid.

eddiebergman:
I fixed the compiler directives to actually by active `cython_directives -> compiler_directives` and it seems I need to convert the `center_range` generator to be cpython to handle big ints for windows. I'm getting this feeling based on the directives I'm not really sure why it's failing but it's my best guess since `wraparound` is now explicitly set to `False` where by [default it is set to `True`](https://cython.readthedocs.io/en/latest/src/userguide/source_files_and_compilation.html#compiler-directives).

ConfigSpace/functional.py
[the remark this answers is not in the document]
mfeurer: Sorry, of course, nevermind.

ConfigSpace/hyperparameters.pyx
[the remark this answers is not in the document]
mfeurer: Thanks for the explanation. This makes sense given that the other function checks for the number of neighbors to generate.

ConfigSpace/hyperparameters.pyx
```diff
@@ -1542,84 +1553,93 @@ cdef class UniformIntegerHyperparameter(IntegerHyperparameter):
...
-        neighbors = []  # type: List[int]
-        cdef int sampled_neighbors = 0
-        _neighbors_as_int = set()  # type: Set[int]
-        cdef long long int_value = self._transform(value)
-        cdef long long new_int_value = 0
-        cdef float new_value = 0.0
-        cdef np.ndarray samples
-        cdef double[:] samples_view
-
-        if self.upper - self.lower <= n_requested:
```
mfeurer, line 1563: Do you think keeping this could reduce the runtime in edge cases? I would bet we can reduce it quite a bit again by re-using this little for-loop

mfeurer (commented):
This now looks great and I'd be happy to merge it. Unfortunately, it's a bit slower generating neighbors for the auto-sklearn search space. Do you think you could re-add the cython annotations to make this fast again?

ConfigSpace/hyperparameters.pyx
eddiebergman: Put it back in :)

ConfigSpace/hyperparameters.pyx
```diff
@@ -1538,88 +1555,129 @@ cdef class UniformIntegerHyperparameter(IntegerHyperparameter):
...
+
+
+        # A truncated normal between 0 and 1, centered on the value with a scale of std.
+        # This will be sampled from and converted to the corresponding int value
+        # OPTIM: Use of clipping insead of scipy.truncnorm
+        # Clipping is faster for generation but more likely to generate clumped values of 0/1
+        # at the boundaries, given a wide enough std and a value close to 0 or 1.
+        # However the overhead of the scipy call seems to outweight this clumping + deduping
+        # procedure in our used benchmark.
+        float_indices = np.clip(rs.normal(value, std, n_requested), 0, 1)
```
eddiebergman, line 1641: This was actually one of the main culprits for the increased time taken.

Scipy's truncnorm is a good bit slower (at least on the time scale we are interested in). It was faster to just clip the normal, as was done before. This may however be detrimental in the case of a value close to the 0-1 boundary and a large enough stddev.

My guess is with a different benchmark, we would end up with a lot of duplicate 0's (clipped values) which means we have to enter the slower loop part below.

ConfigSpace/hyperparameters.pyx
```diff
@@ -1538,88 +1555,129 @@ cdef class UniformIntegerHyperparameter(IntegerHyperparameter):
...
+        # that are not duplicates
+        cdef set seen = {center}
+        cdef set duplicates = set()
+        for i in range(n_requested):
+            v = possible_neighbors[i]
+            if v in neighbors or v == center:
+                duplicates.add(v)
+            else:
+                seen.add(v)
+                neighbors.append(v)
```
eddiebergman, line 1655: Another slight optimization is to not enter the loop for most of the sampled values.

For a brief example, consider the antogonistic case where `possible_neighbors` was `[1, 1, 2, 3, 4, 5]` with the bounds being `(1, 6)`.

* The first `1` would have been added to `neighbors`.
* The second `1` would then enter the recovery case and become a `2`.
* The loop would move on to `2` and find it has also been seen, entering the recovery case.
* and so on...

Now we explicitly make sure to only perform this recovery option on duplicates.

ConfigSpace/util.pyx
```diff
@@ -41,6 +38,9 @@ from ConfigSpace.hyperparameters import CategoricalHyperparameter, \
 import ConfigSpace.c_util
 cimport cython
 
+# OPTIM: To prevent large memory allocations, we place an upperbound on the maximum
+# amount of neighbors to be sampled
+MAX_NEIGHBORHOOD = 10_000
```
eddiebergman, line 43: There was an issue here where the number of requested neighbors for a `UniformIntegerHyperparameter` was set to the full possible range. This would firstly blow up memory as it would try to generate every possible neighbor and secondly, only a fraction of these were used.

This constant can be seen in use later on in this PR.

ConfigSpace/util.pyx
```diff
@@ -185,11 +185,15 @@ def get_one_exchange_neighbourhood(
...
+                    if len(neighbors) == 0:
                         if isinstance(hp, UniformIntegerHyperparameter):
+                            _n_neighbors = min(n_neighbors_per_hp[hp_name], MAX_NEIGHBORHOOD)
                             neighbors = hp.get_neighbors(
-                                value, random,
-                                number=n_neighbors_per_hp[hp_name], std=stdev,
+                                value,
+                                random,
+                                number=_n_neighbors,
+                                std=stdev
```
eddiebergman, line 196: There's a line further below this block `neighbors.pop()`. After fixing an issue with `UniformIntegerHyperparamter::get_num_Neighbors`, this cause the benchmark to fail. There was some parameter with only 3 possible neighbors which must have been sampled a few times, causing it to run out of neighbors during this procedure.

Now it will just generate a new set of neighbors if it has run out from all of the `neighbors.pop`

ConfigSpace/hyperparameters.pyx
```diff
@@ -1538,88 +1555,129 @@ cdef class UniformIntegerHyperparameter(IntegerHyperparameter):
             return False
 
     def get_num_neighbors(self, value = None) -> int:
-        return self.upper - self.lower
+        # If there is a value in the range, then that value is not a neighbor of itself
+        # so we need to remove one
+        if value is not None and self.lower <= value <= self.upper:
+            return self.upper - self.lower - 1
+        else:
+            return self.upper - self.lower
```
eddiebergman, line 1563: I wasn't sure how to handle this but I thought it should act similar to `Categorical` and that the `value` would count as a neighbor. However Categorical just implies it implicitly, regardless of the `value` passed in. Here, I've said that any `value` in the range can not be a neighbor of itself.

eddiebergman:
See above for some more small bits and peices.

As for the timings, it seems to be back to it's original for me:
```

fix_memory_leak:
Average time sampling 100 configurations 0.024013697199999972
Average time retrieving a nearest neighbor 0.003898305980000023
Average time checking one configuration 0.0001229973144186039

master:
Average time sampling 100 configurations 0.023249030113220215
Average time retrieving a nearest neighbor 0.003536670207977295
Average time checking one configuration 0.00012289423312202225
```

On average, `fix_memory_leak` seems about `0.0003` slower but their distributions overlap for me locally.

> RE Windows/wraparound: you could change the flag for that one file?

Let's talk about this on Monday but short answer is I'm not sure how to handle this properly because it appears `int` on mac/linux is an `int64 (long long)` but on windows it's an `int32`. There was no checks for anything of this kind before and `cython` wasn't setting the compiler directives properly before with `wraparound` (so it was silently wrapping without throwing an error).

tldr; I think this is a pre-existing error for large integers on windows and was silently eaten.

mfeurer (approved):

PR #285, opened 2023-01-05
Add support for sigma=0 in normally distributed hyperparameters

nchristensen:
Moving this to separate pull request per #280 

This pull request does mainly two things:
1) It loosens the type restriction to allow floating type values for `mu`. For instance `mu=4.5` might be chosen if one desires four and five to be chosen with equal probability. This required changing `default_value` to be `mu` rounded to the nearest integer  when not otherwise specified. 
2) It allows (or rather handles more correctly) `sigma=0` in NormalIntegerHyperparameter objects. This means the distribution is a Dirac delta function and the only value returned from the distribution is the mean (`mu`). If the `mu` is not an integer then an error is raised during instantiation.

### DIFF
--- a/ConfigSpace/hyperparameters.pyx
+++ b/ConfigSpace/hyperparameters.pyx
@@ -141,7 +141,7 @@ cdef class Hyperparameter(object):
     def has_neighbors(self):
         raise NotImplementedError()
 
-    def get_neighbors(self, value, rs, number, transform = False):
+    def get_neighbors(self, value, rs, number, transform=False):
         raise NotImplementedError()
 
     def get_num_neighbors(self, value):
@@ -297,7 +297,7 @@ cdef class Constant(Hyperparameter):
     def has_neighbors(self) -> bool:
         return False
 
-    def get_num_neighbors(self, value = None) -> int:
+    def get_num_neighbors(self, value=None) -> int:
         return 0
 
     def get_neighbors(self, value: Any, rs: np.random.RandomState, number: int,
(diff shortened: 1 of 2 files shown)


ConfigSpace/hyperparameters.pyx
```diff
@@ -1663,8 +1668,7 @@ cdef class NormalIntegerHyperparameter(IntegerHyperparameter):
     cdef public nfhp
     cdef normalization_constant
 
-
-    def __init__(self, name: str, mu: int, sigma: Union[int, float],
+    def __init__(self, name: str, mu: Union[int, float], sigma: Union[int, float],
```
mfeurer, line 1671: Could you please update the docstring, too?

mfeurer (commented):
Thanks a lot for pulling out this change. Could you please add some comments in this PR on the changes (so it will be easy later to understand this PR) and add some comments (at least where I requested them)?

eddiebergman:
Hi, this PR is out of date and will entirely be so once #346 is merged. As a result, I will close this but thank you for the PR!

However I was still wondering why there's a need for a dirac delta function, given this is equivalent to a `Constant`?

nchristensen:
The idea was to autotune with [ytopt](https://github.com/ytopt-team/ytopt) using a certain fixed values that might change later on. For example, autotuning the multiplication of a 32 x 32 matrix and a 32 x 1000 matrix at first and then later autotuning the multiplication of a 32 x 32 and 32 x 5000 matrix (but using the data from the first case to inform initial model). 

I think I tried using `Constant`, but it either ConfigSpace or ytopt didn't like it when I changed the value of the `Constant` between executions. I ended up taking a different approach (essentially, only tuning the largest operations matrix dimensions).

eddiebergman:
Thanks for the use case! Would it ha e helped if you could easily copy and modify a Configuration space between executions? A lot of ConfigSpace relies on hyperparameter's attributes remaining fixed  but there's no real API for mutating spaces or hyperparameter's which I guess might have helped here

nchristensen:
Yeah, that likely would have been helpful.

PR #297, opened 2023-02-16
Pick the default of categorical hyperparameters based on the weights

### LINKED ISSUE #258: Pick the default of categorical hyperparameters based on the weights
Currently, the categorical hyperparameters pick the first category as the default. This might be a sensible default for categorical hyperparameters with eqaul weights, but in case there are weights attached, we should start we the mode.

herilalaina:
This PR fixes #258 

Set the default value of categorical hyperparameters to the most frequent category if weights are given; otherwise, pick the first category as before.

### DIFF
--- a/ConfigSpace/hyperparameters/categorical.pyx
+++ b/ConfigSpace/hyperparameters/categorical.pyx
@@ -88,6 +88,8 @@ cdef class CategoricalHyperparameter(Hyperparameter):
         self.choices = tuple(choices)
         if weights is not None:
             self.weights = tuple(weights)
+        else:
+            self.weights = None
         self.probabilities = self._get_probabilities(choices=self.choices, weights=weights)
         self.num_choices = len(choices)
         self.choices_vector = list(range(self.num_choices))
@@ -236,7 +238,7 @@ cdef class CategoricalHyperparameter(Hyperparameter):
     def check_default(self, default_value: Union[None, str, float, int]
                       ) -> Union[str, float, int]:
         if default_value is None:
-            return self.choices[np.argmax(self.weights) if hasattr(self, "weights") else 0]
+            return self.choices[np.argmax(self.weights) if self.weights is not None else 0]
         elif self.is_legal(default_value):
             return default_value
         else:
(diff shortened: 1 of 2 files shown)


ConfigSpace/hyperparameters/categorical.pyx
```diff
@@ -236,7 +236,7 @@ cdef class CategoricalHyperparameter(Hyperparameter):
     def check_default(self, default_value: Union[None, str, float, int]
                       ) -> Union[str, float, int]:
         if default_value is None:
-            return self.choices[0]
+            return self.choices[np.argmax(self.weights) if hasattr(self, "weights") else 0]
```
eddiebergman, line 239: Why the `hasattr` I'm on my phone so I can't check the code but I would have expected the attribute to always be there?

eddiebergman (commented):
I suspect a test should fail based on the change, and if not, we should probably have at least one simple test so this behaviour is hammered in :)

ConfigSpace/hyperparameters/categorical.pyx
herilalaina: For now, attribute `self.weights` exists only if `weights` is specified in the initialization. I'll modify that to make this attribute always defined, also so that it can be reused in `__copy__`

herilalaina:
Test cases for checking the default value were added :)

eddiebergman (approved):
Seems good to me!

test/test_hyperparameters.py
```diff
@@ -1838,6 +1838,13 @@ def test_categorical_choices(self):
         ):
             CategoricalHyperparameter('param', ['a', None])
 
+    def test_categorical_default(self):
+        f1 = CategoricalHyperparameter("param", ["a", "b"])
+        f2 = CategoricalHyperparameter("param", ["a", "b"], weights=[0.3, 0.6])
+        f3 = CategoricalHyperparameter("param", ["a", "b"], weights=[0.6, 0.3])
+        self.assertNotEqual(f1.default_value, f2.default_value)
+        self.assertEqual(f1.default_value, f3.default_value)
```
eddiebergman, line 1846: I won't block on this but reading the test alone doesn't illuminate what the intended behavior is meant to be. Either a comment or to say so or a single assert per `fX` could help

mfeurer (approved):
I agree with @eddiebergman that a comment on the test would be great, but besides that this I'm happy about the update.

herilalaina:
Thanks for your feedback :) @mfeurer @eddiebergman

PR #299, opened 2023-02-23
Sampling of neighbors in NormalIntegerHyperparameter

### LINKED ISSUE #298: Sampling of neighbors in NormalIntegerHyperparameter is broken
- In the bounded case, the bounds are not respected
- In the unbounded case, the code crashes
- When everything works, the `number` parameter is not respected (only 1 sample is drawn)

Neonkraft:
This PR resolves #298 
* fix: Fix crash while sampling neighbors in the unbounded case
* fix: Make sampling neighbors in the bounded case respect the bounds
* fix: Make sampling of neighbors respect the `number` parameter
* test: Add tests for the bug


eddiebergman (approved):
I'm quite surprised no test failed due to this bug, like really surprised. Thanks for fixing it :)

PR #301, opened 2023-02-24
Fix float serialization in NormalFloatHyperparameter

### LINKED ISSUE #276: NormalFloatHyperparameter loses bounds when written to JSON
Serializing the following space with a normally distributed float parameter
```python
from ConfigSpace import ConfigurationSpace, Float, Normal
from ConfigSpace.read_and_write import json

cs = ConfigurationSpace(
    space={
        "p": Float(
            "p",
            bounds=(1e-5, 1e-1),
            default=0.01,
            log=True,
            distribution=Normal(1.0, 0.6),
        ),
    },
)

print(json.write(cs))
```

produces the output
```json
{
  "hyperparameters": [
    {
      "name": "p",
      "type": "normal_float",
      "log": true,
      "mu": 1.0,
      "sigma": 0.6,
      "default": 0.01
    }
  ],
  "conditions": [],
  "forbiddens": [],
  "python_module_version": "0.6.0",
  "json_format_version": 0.4
}
```

which I expected to contain the bounds as well, but they seem to get lost during JSON serialization.

A test case that I'd expect to pass and cover this issue could look like
```python
cs1 = ConfigurationSpace(
    space={
        "p": Float(
            "p",
            bounds=(1e-5, 1e-1),
            default=0.01,
            log=True,
            distribution=Normal(1.0, 0.6),
        ),
    },
)
cs2 = json.read(json.write(cs1))
assert cs1 == cs2
```

Neonkraft:
Resolves #276

### DIFF
--- a/test/test_util.py
+++ b/test/test_util.py
@@ -34,8 +34,10 @@ import numpy as np
 
 from ConfigSpace import Configuration, ConfigurationSpace, UniformIntegerHyperparameter, \
     UniformFloatHyperparameter, CategoricalHyperparameter, Constant, OrdinalHyperparameter, \
-    EqualsCondition, AndConjunction, OrConjunction, LessThanCondition, GreaterThanCondition
+    EqualsCondition, AndConjunction, OrConjunction, LessThanCondition, GreaterThanCondition, \
+    Float, Normal
 from ConfigSpace.read_and_write.pcs import read
+from ConfigSpace.read_and_write import json
 from ConfigSpace.util import impute_inactive_values, get_random_neighbor, \
     get_one_exchange_neighbourhood, deactivate_inactive_hyperparameters, fix_types, generate_grid
 import ConfigSpace.c_util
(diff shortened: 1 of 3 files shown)


test/test_util.py
```diff
@@ -531,3 +533,10 @@ def test_generate_grid(self):
         # Check 1st and last generated configurations completely:
         self.assertEqual(generated_grid[0].get_dictionary()['float1'], -1.0)
         self.assertEqual(generated_grid[-1].get_dictionary()['float1'], 1.0)
+
+    def test_hyperparameter_json_serialization(self):
```
mfeurer, line 537: I think this is the wrong file for such a test and it should rather go into `test/read_and_write/test_json.py`.

mfeurer (changes requested):
Looks almost good, but I think there's one test that needs to go to a different file.

test/test_util.py
Neonkraft: Done.

eddiebergman:
I guess this is superseeded by #308 and I can close this?

Neonkraft:
> I guess this is superseeded by #308 and I can close this?

Yes.

PR #314, opened 2023-04-18
Fixes grid generation with forbidden clauses #312

### LINKED ISSUE #312: Grid generation does not respect forbidden clauses
It seems that `generate_grid` does not respect forbidden clauses: if there is a forbidden clause an exception is raised because it is trying to instantiate a Configuration with an invalid set of values.

Reproducible code (v0.6.1):
```python
from ConfigSpace.util import generate_grid
from ConfigSpace import ConfigurationSpace, Categorical, ForbiddenEqualsClause, ForbiddenAndConjunction

cs = ConfigurationSpace(space={
    'a': Categorical('a', ['A', 'B', 'C']),
    'b': Categorical('b', ['E', 'F']),
})
cs.add_forbidden_clause(ForbiddenAndConjunction(  # Forbid b == F iff a == C
    ForbiddenEqualsClause(cs['a'], 'C'),
    ForbiddenEqualsClause(cs['b'], 'F'),
))

# Sampling does indeed respect forbidden clauses: there are 5 possible configurations
seen = set()
for config in cs.sample_configuration(100):
    seen.add(tuple(config.get_array()))
assert len(seen) == 5

# Raises AssertionError and ForbiddenValueError
configs = generate_grid(cs)
assert len(configs) == 5
```

The AssertionError comes from [here](https://github.com/automl/ConfigSpace/blob/main/ConfigSpace/util.pyx#L641) in the generate_grid function. It expects some other exception to be raised, which is not the case, so it raises an AssertionError.

The real underlying error though is the ForbiddenValueError raised when creating a Configuration and checking `is_valid_configuration`.

A quick fix would just be to skip any Configuration created that raises a ForbiddenValueError. Better would be to detect this condition upfront and prevent the creation in the first place.

jbussemaker:
This pull request should fix #312 by simply rejecting generated Configurations that violate any forbidden clause

### DIFF
--- a/ConfigSpace/util.pyx
+++ b/ConfigSpace/util.pyx
@@ -642,6 +642,12 @@ def generate_grid(configuration_space: ConfigurationSpace,
         try:
             grid_point = Configuration(configuration_space, unchecked_grid_pts[0])
             checked_grid_pts.append(grid_point)
+
+        # When creating a configuration that violates a forbidden clause, simply skip it
+        except ForbiddenValueError:
+            unchecked_grid_pts.popleft()
+            continue
+
         except ValueError as e:
             assert (str(e)[:23] == "Active hyperparameter '" and
                     str(e)[-16:] == "' not specified!"), \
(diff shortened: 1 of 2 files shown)


eddiebergman (approved):

eddiebergman:
Heyo, tests seem to fail, seems like a seeding issue?

eddiebergman:
Many thanks for the contributions!!

Issue #315, opened 2023-04-20
Documentation GitHub Icon Points to Wrong Repository 


LennartPurucker:
Heyho, the GitHub icon on https://automl.github.io/ConfigSpace/main/api/index.html points to https://github.com/automl/automl_sphinx_theme.

mfeurer:
Thanks a lot for reporting. Would you like to create a PR to solve this issue?

PR #316, opened 2023-04-27
fix(doc): fix link in Github icon on the docs page

Neonkraft:


Neonkraft:
This PR resolves #315

PR #329, opened 2023-07-04
fix: fix illegal default values in bounded hyperparameters

### LINKED ISSUE #310: NormalIntegerHyperparameter and NormalFloatHyperparameter accepts illegal default values
```
from ConfigSpace import Normal, Integer, Float
int_hyper = Integer("p", bounds=(0.0, 2.0), default=5., distribution=Normal(1.0, 0.2))
float_hyper =  Float("p", bounds=(0.0, 2.0), default=5., distribution=Normal(1.0, 0))
```
Both these lines should throw exceptions since the default values are outside the bounds, but the code runs just fine.

Neonkraft:
Resolves #310

### DIFF
--- a/ConfigSpace/hyperparameters/normal_float.pyx
+++ b/ConfigSpace/hyperparameters/normal_float.pyx
@@ -207,7 +207,7 @@ cdef class NormalFloatHyperparameter(FloatHyperparameter):
                                            q=q_int, log=self.log)
 
     def is_legal(self, value: Union[float]) -> bool:
-        return (isinstance(value, float) or isinstance(value, int))  and \
+        return (isinstance(value, (float, int, np.number)))  and \
                (self.lower is None or value >= self.lower) and \
                (self.upper is None or value <= self.upper)
 
--- a/ConfigSpace/hyperparameters/normal_float.pyx
+++ b/ConfigSpace/hyperparameters/normal_float.pyx
@@ -207,7 +207,9 @@ cdef class NormalFloatHyperparameter(FloatHyperparameter):
                                            q=q_int, log=self.log)
 
     def is_legal(self, value: Union[float]) -> bool:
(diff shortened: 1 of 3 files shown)


ConfigSpace/hyperparameters/normal_float.pyx
```diff
@@ -207,7 +207,9 @@ cdef class NormalFloatHyperparameter(FloatHyperparameter):
                                            q=q_int, log=self.log)
 
     def is_legal(self, value: Union[float]) -> bool:
-        return isinstance(value, float) or isinstance(value, int)
+        return (isinstance(value, float) or isinstance(value, int))  and \
```
eddiebergman, line 210: Can we make this `isinstance(value, (float, int, np.number))`?

ConfigSpace/hyperparameters/normal_integer.pyx
```diff
@@ -195,7 +195,9 @@ cdef class NormalIntegerHyperparameter(IntegerHyperparameter):
                                             q=self.q, log=self.log, meta=self.meta)
 
     def is_legal(self, value: int) -> bool:
-        return isinstance(value, (int, np.int32, np.int64))
+        return (isinstance(value, (int, np.int32, np.int64))) and \
```
eddiebergman, line 198: Likewise for here `isinstance(value, (int, np.integer))`

eddiebergman (changes requested):

ConfigSpace/hyperparameters/normal_float.pyx
Neonkraft: Done.

ConfigSpace/hyperparameters/normal_integer.pyx
Neonkraft: Done.

eddiebergman (approved):

PR #332, opened 2023-07-07
Remove hyperparameter in a confiugration space

### LINKED ISSUE #267: Create easy way to remove hyperparameters
YAHPOBench has the task id in it's configuration as a hyperparamter which is not something you would like any optimizer to know about. Should be an easy way to remove a hyperparamter by name.

If this violates some conditional where the remove hyperparamter is what is being condition on, i.e. `a` is active when `b` is `x` and we remove `b` from the space. Then we need to remove this condition too. Likewise for forbidden clauses.

herilalaina:
Closes #267.

This PR creates a new feature that allows removing a hyperparameter in a configuration space. The implementation is built upon the initial code given by @eddiebergman in the mentioned issue, mainly also to remove the conditions and forbidden clauses referenced by the hyperparameter.

### DIFF
--- a/ConfigSpace/conditions.pyx
+++ b/ConfigSpace/conditions.pyx
@@ -74,6 +74,9 @@ cdef class ConditionComponent(object):
     def evaluate_vector(self, instantiated_vector):
         return bool(self._evaluate_vector(instantiated_vector))
 
+    def get_referenced_hyperparameters(self) -> List[Hyperparameter]:
+        pass
+
     cdef int _evaluate_vector(self, np.ndarray value):
         pass
 
@@ -146,6 +149,9 @@ cdef class AbstractCondition(ConditionComponent):
                  ) -> bool:
         hp_name = self.parent.name
         return self._evaluate(instantiated_parent_hyperparameter[hp_name])
+
+    def get_referenced_hyperparameters(self) -> List[Hyperparameter]:
+        return self.get_children() + self.get_parents()
 
     cdef int _evaluate_vector(self, np.ndarray instantiated_vector):
         if self.parent_vector_id is None:
@@ -522,6 +528,9 @@ cdef class AbstractConjunction(ConditionComponent):
             if c1 != c2:
                 raise ValueError("All Conjunctions and Conditions must have "
                                  "the same child.")
+
+    def get_referenced_hyperparameters(self) -> List[Hyperparameter]:
+        return sum([cond.get_referenced_hyperparameters() for cond in self.components], [])
 
     def __eq__(self, other: Any) -> bool:
         """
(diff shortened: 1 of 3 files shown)


ConfigSpace/conditions.pyx
```diff
@@ -523,6 +529,9 @@ cdef class AbstractConjunction(ConditionComponent):
                 raise ValueError("All Conjunctions and Conditions must have "
                                  "the same child.")
 
+    def get_referenced_hyperparameters(self) -> List[Hyperparameter]:
+        return sum([cond.get_referenced_hyperparameters() for cond in self.components], [])
```
eddiebergman, line 533: Nice pattern, didn't know you could do this

ConfigSpace/util.py
```diff
@@ -708,3 +708,68 @@ def get_cartesian_product(value_sets: list[tuple], hp_names: list[str]) -> list[
...
+        if hp_to_remove not in cond.get_referenced_hyperparameters()
+    ]
+    forbiddens = [
+        forbidden
+        for forbidden in configuration_space.get_forbiddens()
+        if hp_to_remove not in forbidden.get_referenced_hyperparameters()
+    ]
+
+    if isinstance(configuration_space.random, np.random.RandomState):
+        new_seed = configuration_space.random.randint(2**32 - 1)
```
eddiebergman, line 752: We should probably copy the `RandomState`, rather then create a new one: https://stackoverflow.com/a/63875402/5332072

eddiebergman (changes requested):
Might be good to have tests for this, I imagine there are possible edge cases where this fails quite drastically

ConfigSpace/util.py
herilalaina: Very interesting.  I have always thought it is not possible to copy RandomState.

herilalaina:
I have added a few tests to cover different cases. Let me know if you have any specific scenarios that should be included.

herilalaina:
I have added the requested tests already. @eddiebergman is there anything I missed again? :)

eddiebergman:
Maybe, it just completely slipped my stack. I'll take a look next week when I'm back!

eddiebergman:
Hi @herilalaina and @simonprovost,

There is an issue in that using Cython 3 to compile ConfigSpace doesn't work anymore as it now expects more rigid typing (of which there are quite a few inconsistencies).

I will manually edit and merge this PR once the PR at #346 is done

eddiebergman:
Will keep this open in the time being, #346 has been merged

eddiebergman:
I would wait for what I assume to be @mfeurer to do this, I assume he will give you full access rights.

thijssnelleman:
Closing this as now #411 has been merged

PR #345, opened 2023-12-11
Added TypedConfigurationSpace

MischaPanch:
I provided two ways of creating typed configuration spaces - one binding a class to an output type, and another where the type constructor is simply passed at init. The second option requires less code, doesn't use generics, and many users might prefer it. However, the first is more explicit, allows for better discoverability and might be prefered by users who have to deal with mutliple config spaces

I have also fixed some types and slightly enhanced input validation. Previously, passing `name={..},  space=some_space` would have simply ignored that space was passed at all

### DIFF
--- a/ConfigSpace/api/types/categorical.py
+++ b/ConfigSpace/api/types/categorical.py
@@ -119,7 +119,7 @@ def Categorical(
         :py:class:`CategoricalHyperparameter`.
 
     meta: dict | None = None
-        Any additional meta information you would like to store along with the hyperparamter.
+        Any additional meta information you would like to store along with the hyperparameter.
     """
     if ordered and weights is not None:
         raise ValueError("Can't apply `weights` to `ordered` Categorical")
--- a/ConfigSpace/configuration_space.py
+++ b/ConfigSpace/configuration_space.py
@@ -188,7 +188,7 @@ class ConfigurationSpace(Mapping[str, Hyperparameter]):
         self.name = name
         self.meta = meta
 
-        # NOTE: The idx of a hyperparamter is tied to its order in _hyperparamters
+        # NOTE: The idx of a hyperparamter is tied to its order in _hyperparameters
         # Having three variables to keep track of this seems excessive
         self._hyperparameters: OrderedDict[str, Hyperparameter] = OrderedDict()
         self._hyperparameter_idx: dict[str, int] = {}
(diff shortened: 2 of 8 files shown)


MischaPanch:
@eddiebergman Here the PR - better late than never :). Would be happy if it could be merged in the foreseeable future, as I'd like to use configspace for our projects. Meanwhile, I can of course just use this extension in our own codebase

eddiebergman:
Hi @MischaPanch, I won't be able to properly review it until I'm back next week but I've set a reminder for it! Many thanks for the PR :)

eddiebergman:
Also as a side note, there'll likely be build issues due to needing to update the types to comply with the newer version of CYthon, which will likely take a few days to complete as well, before we can make it publically available on pypi. It's some technical debt I've been ignoring for a while but it's a good reason to update it :)

MischaPanch:
> Hi @MischaPanch, I won't be able to properly review it until I'm back next week but I've set a reminder for it! Many thanks for the PR :)

No problem at all, take your time!

eddiebergman:
Hi @herilalaina and @simonprovost,

There is an issue in that using Cython 3 to compile ConfigSpace doesn't work anymore as it now expects more rigid typing (of which there are quite a few inconsistencies).

You might need to rebase once #346 is done, from what I checked, the only thing you'll need to actively fix is then the `unittest` test case for what you have. I will do a review once the refactor is done anywho

eddiebergman:
Hi @MischaPanch,

It's taken a while but there was a major refactor in #346 which means this PR would likely need to be updated. I'm also happy to go ahead and try implementing the feature in the refactored version given this PR as a base and inspiration!

FYI: I've tested and everythings been deprecated properly such that most old ConfigSpace tests and all SMAC's unittests pass with #346 installed. However you might want to know that #346 hard removed `"q"` quantization of hyperparameters and unbounded hyperparameters, everything else is soft-deprecated and should function as before.

I think this should be closed and a new PR re-opened, but I will leave it to you if you'd like to continue this work!

Also, sorry for the delay on this but hopefully you like the new ConfigSpace :)

mfeurer:
Closing this due to inactivity. @MischaPanch if you are still interested, please let us know.

MischaPanch:
Hi, sorry for the lack of response. Feel free to do with this code as you see fit :). I don't need this to be merged

PR #350, opened 2024-02-06
Quantization type hint bug fix

### LINKED ISSUE #349: Inconsistent typing on quantization parameter for Float hyperparameters
The docstring says it supports `float` but the typing hints only say int, even though the base classes it feeds into support float.  Seems like a small bug to me.

image

bpkroth:
Fixes #349

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -70,5 +70,8 @@ prof/
 # VSCode
 .vscode
 
+# Virtual environments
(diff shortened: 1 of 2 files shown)


eddiebergman:
Many thanks :)

eddiebergman (approved):

PR #361, opened 2024-05-13
Bugfix for problematic typecheck when creating Integer

Ultrahero:
When creating a UniformIntHyperparameter, the default_value is parsed as an int to the uniformFloatHyperparameter, which causes the type_check in line 62 to fail.

This is obviously a bug. This type conversion should fix the bug.


eddiebergman:
Hi @Ultrahero ,

Appreciate the PR but in #346 we've fixed a lot of bugs and just waiting for some more review before releasing. I will close this PR as a result but thank you again for the PR!

PR #366, opened 2024-07-03
fix(hyperparameters): fix #362

### LINKED ISSUE #362: TypeError: Expected float, got numpy.float64
There is an issue with this file:

`ConfigSpace/hyperparameters/uniform_float.pyx:124`: TypeError: Expected float, got numpy.float64 

It occurs when attempting to create a `UniformFloatHyperparameter` with or without passing a default value.

**I think the issue is casting to float before rounding instead of doing the opposite.** 

Original code: `default_value = np.round(float(default_value), 10)`

Suggested fix: `default_value = float(np.round(default_value, 10))`

Python version: 3.11.9 
configspace: 0.7.2 

Note: the error does not exist when using Python v3.9.19 with configspace v0.6.1

Crylab:
assign a python float as result when checking or the default value, fixes #362


eddiebergman:
We got rid of Cython completely in #346 and the problem should be fixed. Many thanks for the PR though!

Issue #372, opened 2024-07-15
Bug: ForbiddenEqualsRelation does not work


mdorier:
Example program:

```
from ConfigSpace import (
            Configuration,
            ConfigurationSpace,
            Integer,
            ForbiddenEqualsRelation)

  cs = ConfigurationSpace()
  cs.add(Integer("x", (1, 3)))
  cs.add(Integer("y", (0, 2)))
  cs.add(ForbiddenEqualsRelation(cs["y"], cs["x"]))

  print(cs)

  print(cs.sample_configuration())
```

This program will sometimes output a configuration where `x == y`.

mdorier:
The problem is here: https://github.com/automl/ConfigSpace/blob/e03e2604ff6d5e485c7ccb30aa1905e51b7eaedb/src/ConfigSpace/forbidden.py#L689

PR #373, opened 2024-07-15
Fix ForbiddenEqualsRelation

mdorier:
Fixes https://github.com/automl/ConfigSpace/issues/372


eddiebergman:
Ahhh, I'm sorry about this one. Stray casualty of the recent 1.0 update. Surprised there wasn't a test that caught this. I will merge and add a test for it tomorrow! Will also release a 1.0.1

PR #406, opened 2025-06-13
fix: shuffle to undo unique sort

### LINKED ISSUE #405: Neighbour Values of Integer Hyperparameter are not Symmetric
Hi,

When running the following code:

```python
from ConfigSpace import UniformIntegerHyperparameter
import numpy as np

rs = np.random.RandomState(0)

hp = UniformIntegerHyperparameter("a", lower=0, upper=2048, log=False)
neighbors = hp.neighbors_values(1024, n=100, seed=rs)
print(neighbors)
```

I get as a result:

```
[100 113 176 192 212 231 269 304 317 325 328 337 345 356 361 362 363 367
 375 380 391 393 394 413 421 423 428 435 440 442 457 461 472 477 485 489
 494 503 511 518 522 526 537 539 540 543 545 547 550 554 562 571 573 576
 581 585 590 594 596 600 605 614 617 620 624 632 637 650 652 657 659 660
 667 671 674 685 691 695 708 711 712 718 719 720 721 722 725 727 729 735
 739 740 741 742 743 744 745 748 752 757]
```

I would suspect that a neighbourhood should be symmetric around the value of 1024 since the documentation of the method states "Sample neighbours of a value". Is this a bug or am I misunderstanding something? 

ConfigSpace Version: 1.2.1
Python Version: 3.10.11

Thanks for the help!

eddiebergman:
* Closes #405 

@LukasFehring Anyone going to take over on `ConfigSpace` from Hannover side? I'm not supported to really maintain `ConfigSpace` as I've left AutoML.org

### DIFF
--- a/src/ConfigSpace/hyperparameters/distributions.py
+++ b/src/ConfigSpace/hyperparameters/distributions.py
@@ -181,6 +181,11 @@ def quantized_neighborhood(
         uniq = np.unique(candidates_quantized)
         new_uniq = np.setdiff1d(uniq, neighbors[:offset], assume_unique=True)
 
+        # Unfortunately, uniq sorts the values, so we lose the "centered"'ness, and
+        # `uniq` and `new_uniq` are now left-biased samples. We thus need to undo this
+        # sort operation.
+        seed.shuffle(new_uniq)
+
         n_new_unique = len(new_uniq)
         neighbors[offset : offset + n_new_unique] = new_uniq
         offset += n_new_unique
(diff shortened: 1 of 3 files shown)


eddiebergman:
It's mergable, you can view the issue it's derived from and that it fixed the users issue.

Essentially, the sampling procedure would require doing a check "have we already sampled this value", which is much cheaper to check when sorted. However this sorting ended up biasing the sample, as we returned the first `n` samples, i.e. the first `n` samples according to order, instead of a _random_ `n` samples around the distribution

thijssnelleman:
Alright, sounds good to me. @mfeurer I think with your approval this can be merged.

PR #414, opened 2025-11-18
configspace.update for hyperparameter settings

thijssnelleman:
https://github.com/automl/ConfigSpace/issues/412

This is a proposed solution to this issue; Hyperparameters have properties that are not updated by changing the values. By adapting the updating functionality, it first tests if there already is a value present, and if so, verifies what needs to be recalculated.

1. Assumes that in the initialisation all values are set exactly once, otherwise the update will be triggered and may cause issues when properties are accessed that do not exist yet
2. Shares a lot of code with that found in the init function. Would be better if this would be settled in one function rather than spread over two
3. Needs to be adapted for each HP type. Need to find out if we can do this more generally (For example, what can be done in the parent Hyperparameter class? Not everything but maybe something like default value?)

Have written a partial pytest to verify the behaviour, seems to work just fine with this suggested solution


eddiebergman (commented):
Might be a lot easier to just reconstruct the whole hyperparameter, and then steal it's values. This means you don't have to duplicate validation, and you can use this method with minor adjustments on everywhere.

Read up on [dataclasses](https://docs.python.org/3/library/dataclasses.html). There is a method signature `.replace() -> Self` which will basically do this, but it returns a new instance.
> *Just take special note that we define a custom `__init__` for dataclasses, which is not normally what you should do. This means you don't get all their benefits. They exist to be backwards compatbile with the public API*

```python
def __setattr__(self, name: str, value: Any):
    init_params: dict[str, Any] = # get init params
    if name not in init_params:
        raise ValueError("Can't set attribute {name}, must be one passed to init.") # Something better error message than this
    init_params[name] = value
    
    # This raises if invalid
    new_instance = self.__class__(**init_params)
    self.__dict__ = new_instance.__dict__ # Unsure if this works exactly
```

---

thijssnelleman:
@eddiebergman 

> That is actually a good idea, although it could be inefficient in some cases. I went with a slightly different strategy; the **setattr** has now a different workflow if a value is **not** being set from **init**. The check is quite ugly, but works. Instead of using replace, I am just calling self.**init** again. Although current tests are basics, they are passing, wondering what @eddiebergman thinks of this solution.

Does this solutions stray too far from your initial idea?

thijssnelleman:
@mfeurer Reminder for this PR

src/ConfigSpace/hyperparameters/hyperparameter.py
```diff
@@ -132,11 +133,28 @@ def __init__(
...
-                f" hyperparamter '{self.name}'.",
+                f" hyperparameter '{self.name}'.",
             )
 
         self._normalized_default_value = self.to_vector(self.default_value)
 
+    def __setattr__(self, name: str, value: Any):
+        """Check if attribute can be set on HP, and reinitialises the class if so."""
+        # NOTE: The following check is 'ugly', but it works...
+        if inspect.stack()[1][3] != '__init__':  # This should be only executed on update, not init
```
karibbov: I don't think this is a particularly 'ugly' check. However, We should note for documentation that this check would satisfy from within any function named `__init__`. Checking the caller class is possible, but might be too complex and unlikely use case to do it here. (See: https://stackoverflow.com/a/2220759/8889365)
Maybe we can even directly compare the `__code__` objects or some properties of them for equality of caller and `self.__init__`. My opinion would be to just add more comments to why this is currently the best tradeoff for our use case.

src/ConfigSpace/hyperparameters/hyperparameter.py
```diff
@@ -132,11 +133,28 @@ def __init__(
...
+            # Extract all editable attributes
+            init_params: tuple[str] = self.__init__.__code__.co_varnames[:self.__init__.__code__.co_argcount]
+
+            if name not in init_params or not hasattr(self, name):
+                raise ValueError("Can't set attribute {name}, must be one passed to init.") # Something better error message than this
```
karibbov:
```suggestion
                raise ValueError(f"Can only set parameters passed to self.__init__  {name} is not one of: {init_params}")
```

src/ConfigSpace/hyperparameters/hyperparameter.py
```diff
@@ -132,11 +133,28 @@ def __init__(
...
+
+            init_params = {key: self.__dict__[key] for key in init_params if hasattr(self, key)}  # This will break if the parameter is not saved under its passed name
```
karibbov: Perhaps, we can return a more explicit error message here, so a user understands what exactly went wrong. Wrote a rough example, but perhaps we can find a better message. 

```suggestion
            try:
                init_params = {key: self.__dict__[key] for key in init_params if hasattr(self, key)}  # This will break if the parameter is not saved under its passed name
            except KeyError as e:
                raise KeyError(f"You are seeing this message because the class {self.__class__.__name__} does not define an instance attribute with the same name as the class constructor parameter. To update Hyperparameter attributes after initialization, either modify the __init__ function of the class to define the attribute with the same name or call hp_instance.__init__(**new_init_parameters) to reset the parameters") from e
```

src/ConfigSpace/hyperparameters/hyperparameter.py
```diff
@@ -132,11 +133,28 @@ def __init__(
...
+            init_params[name] = value  # Place the update value
+
+            self.__init__(**init_params)  # Reinitialise
```
karibbov, line 161: We should document somewhere that the `__init__` functions for hyperparameter classes must be idempotent and define attributes for every init param with the same name and value. Additionally, we should document that no new attribute definitions or updates are allowed after initialization, except reinitializing the class, and the `self.__setattr__` is a shorthand that ensures this.

test/test_hyperparameters.py
```diff
@@ -3080,3 +3080,94 @@ def test_arbitrary_object_allowed_in_categorical_ordinal(
...
+    for value in sample:
+        assert -0.1 <= value["b"] <= 0.1
+
+    # Test illegal changes
+    with pytest.raises(ValueError):
+        space["b"].upper = -0.11  # lower than lower
+    with pytest.raises(ValueError):
+        space["b"].lower = 0.11  # higher than upper
+    with pytest.raises(ValueError):
+        space["b"].default_value = -10.0  # Out of bounds
```
karibbov, line 3146: Can we add more cases for illegal changes, e.g. not an init parameter, attribute not defined by the same name, an attempt to define a new attribute

karibbov (changes requested):
Summary:
This PR:
1. Prevents direct (assignment) attribute setting from anywhere except from within any function named `__init__`.
2. Then allows only the attributes that has the same name as an initialization parameter, by reinitializing from those attributes. This implies all the initialization parameters must be saved without modification at every `HP.__init__` implementation. This is not as strict of a condition as it seems, since all HPs are dataclasses. However I think we should document it somewhere visible since we always have a custom `__init__` function.
# Suggestion
1. Mainly documetation (see comments)
2. Can we include an example test for all the other init parameters (attributes) being the same before and after an attribute is reset? This is to catch the case when some `HP.__init__` saves a modified version of the init parameter as an attribute.

test/test_hyperparameters.py
thijssnelleman: What do you mean with "attribute not defined by the same name"? I cannot differentiate this from attempting to define new attribute

src/ConfigSpace/hyperparameters/hyperparameter.py
thijssnelleman: Updated the comment, feel free to hit resolve if you're happy with it!

test/test_hyperparameters.py
karibbov: Ah yeah, true

src/ConfigSpace/hyperparameters/hyperparameter.py
thijssnelleman: I have added documentation on this in the docs/reference/hyperparameters.md, lets continue this discussion there.

src/ConfigSpace/hyperparameters/hyperparameter.py
karibbov: Maybe we add the link to the stackoverflow to the comment as well, so a future maintainer has more context? otherwise we can resolve

karibbov: I just realized the cleanest way to work around this would be to block all `self.__setattr__` calls and use `object.__setattr__` in `__init__` intead. Do you think it's a better solution?

thijssnelleman: That would then no longer allow people to actually set attributes, which is kind of the point of this PR (And its solution is to re-initialise the object to save many lines of code and headache)

karibbov: Oh no, I mean removing just the if check and then replacing all current uses of attribute assignments inside `self.__init__` to use `object.__setattr__`. This way no matter where we call attribute assignment we would know that it's re-initializing.

test/test_configuration_space.py
```diff
@@ -1338,3 +1339,171 @@ def test_configuration_space_can_be_made_with_sequence_of_hyperparameters() -> N
...
+            EqualsCondition(space["d"], space["c"], 2),
+        ),
+    )
+    space.add(ForbiddenEqualsClause(space["e"], 0.125))
+
+    space.update(space["a"], 25, "lower")
+    space.update(space["e"], 0.66, "upper")
+    # Check that this is still a valid configuration
+    assert (
+        space._check_configuration_rigorous(
```
karibbov, line 1483: Would these checks always produce an error when we shrink the bounds? I. e. they might produce valid vectors but not cover the whole range in that vector space. Would checking if the new bounds match 0 and 1 be a more explicit test? or if the vector values of the new bounds match the vector values of the old bounds? I don't know if this is possible to query from the vector_dist, but would be closer to a definitive test if we could

thijssnelleman: We have previously tested that updating the properties actually sets them to their new values. What matters after is that the re-initialisation of the object is done correctly e.g. the vectors are updated for the checks. 

We could also check the vector directly, although I would not know how to naturally do that. But given that this is the actual usage that needs to pass I figured this makes the most sense.

Issue #266, opened 2022-08-19
Create easy way to copy a ConfigurationSpace


eddiebergman:
Would be useful in the case of wanting to copy and then modify but retain the original.

thijssnelleman:
Could be done by re-feeding all attributes into the constructor?

PR #423, opened 2026-02-18
Creating copy function on configuration space

thijssnelleman:
Adding a __copy__ method on configuration space as requested in #266


thijssnelleman:
Not sure if the `ConfigurationSpace.__eq__` method is strong enough to serve as a proper test of this solution

Issue #380, opened 2024-07-19
[question] Why not allow multiple conditions on the same parameter?


mdorier:
When I try to add multiple conditions that affect the same hyperparameter, ConfigSpace complains that adding multiple conditions is ambiguous. For example:

```
from ConfigSpace import ConfigurationSpace, EqualsCondition, Integer
cs = ConfigurationSpace()
cs.add(Integer("x", (1, 10), default=1))
cs.add(Integer("y", (1, 10), default=1))
cs.add(Integer("z", (1, 10), default=1))
cs.add(EqualsCondition(cs["z"], cs["x"], 3))
cs.add(EqualsCondition(cs["z"], cs["y"], 3))
```

Gives me:
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/ubuntu/Code/spack/var/spack/environments/bedrock-env/.spack-env/view/lib/python3.11/site-packages/ConfigSpace/configuration_space.py", line 330, in add
    self._dag.add_condition(condition)
  File "/home/ubuntu/Code/spack/var/spack/environments/bedrock-env/.spack-env/view/lib/python3.11/site-packages/ConfigSpace/_condition_tree.py", line 681, in add_condition
    raise AmbiguousConditionError(
ConfigSpace.exceptions.AmbiguousConditionError: Adding a second parent condition for a for a hyperparameter is ambiguous and therefore forbidden. Use an `OrConjunction` or `AndConjunction` to combine conditions instead.
Already inserted: z | x == 3
New one: z | y == 3
```

I don't understand why it would be ambiguous. When adding multiple conditions, it would seem obvious to me that they are in conjunction with each other, not in disjonction (if anything, simply because conditions that affect distinct parameters are already in conjunction with each other).

More practically speaking, the reason this is annoying in my use-case is that, as part of a library, I have a function that returns a `ConfigurationSpace` for the parameters available to the library. The user then has a chance to further constrain this space by adding conditions on parameters. But some of the parameters already have some conditions attached to them by the library itself. 

As is, ConfigSpace doesn't seem to provide an API to remove a previously defined condition and add it back as part of an `AndConjunction`, or add more conditions to an existing `AndConjunction`.

mdorier:
I edited the title to make it a feature request. I've been trying to get around this constraint in my code but it's becoming unmanageable. I really need a way to add multiple conditions to the same hyperparameter from different points in the code (which may be from potentially different libraries).

[...]

PR #424, opened 2026-02-19
Creating proposed solution to multiple conditions problem

thijssnelleman:
Issue from #380 where adding a condition on a parameter which already has a conditions yields an ambiguity error. In this proposed solution, we would synthesise an AND conjunction instead.


thijssnelleman:
The proposed solution has been confirmed to cover the issue #380 by mdorier (reporter)

thijssnelleman:
@mfeurer I have actually ran into this issue elsewhere as well, prioritising this PR would be helpful for me

Issue #207, opened 2021-12-14
Improve grid generation


mfeurer:
The method added in #142 has two drawbacks that can become a limiting factor:

1. It generates all configurations at once. Instead it should be a generator.
2. It does not delegate generating the legal values of a hyperparameter to the respective hyperparameter class. This will make it not pick up on new hyperparameters which can be added.

thijssnelleman:
@mfeurer Is this issue still applicable?

mfeurer:
I think so.

PR #426, opened 2026-02-20
Improve grid generation for #207

thijssnelleman:
Based on the description of #207, I have revamped grid generator to work completely with generators instead of in memory grids.

1. Each hyperparameter gets a constructed tuple or generator (Cannot reduce categorical/ordinal/constant in memory space, their impact should be very limited)
2. Generators are fed into a itertools.product to do a cartesian product
3. Conditional parameters get added a "None" value to their possible values, to represent inactivity, which are filtered out when constructing the configuration from the sample
4. Samples are tested and discarded when invalid.

Biggest issue: I did a timing test based on the pytests, measuring simply with time.time.

The first test yielded the results;
New: 0.008675813674926758 (Slight improvement)
Old: 0.00872802734375

The second, with some conditionals:
New: 0.017199277877807617 (Order of magnitude slower)
Old: 0.008735895156860352

This could be terrible for large search spaces. One thing I (over) simplified was to simply generate the entire grid by the generator and discarding the invalid configurations (i.e. no 'smart' checking for active/inactive parameters). This means a lot of invalid configurations are generated in the second case: 558 / 576 are invalid (96.9%).

 I hoped that the check would be fast enough to mitigate this slow down, but perhaps this is not the case.


thijssnelleman:
The speed issue has now been fixed; conditional parameters are only searched when relevant. Speed test results between the **original** method and this PR:


Simple test (no conditions):
New Time:  0.008033990859985352 (slight improvement)
Old time:    0.008356094360351562

Large test (multiple conditionals)
New Time:   0.0005440711975097656 (Order of magnitude faster than old method, 
Old time:     0.001583099365234375

